Issue #301, opened 2026-06-13
Bind independent safety-controller execution evidence into the audit chain (embodied agents, follow-up to #293)


carloshvp:
## Context

The industrial-embodied-ai example (agentrust-io/examples#16, refined in #18)
deliberately leaves one boundary open and documents it under "Evidence
boundaries": the audit bundle binds request hashes and authorization decisions,
but the controller outcome is recorded only as a client-observed value, not as
independent execution evidence. The README states plainly that binding
independent controller evidence is a follow-up design question rather than
something the example should silently invent.

#293 closed the adjacent plumbing gap: the proxy now populates
`response_payload_hash` for the post-scan response it forwards. That is necessary
but not sufficient for embodied agents, and this issue is about the part #293
does not cover.

## The gap

For an embodied agent, the authoritative outcome is decided by an independent
functional-safety controller that sits on the far side of the tool boundary.
cMCP can hash the response it forwards, but that response is the tool server's
self-report. The audit chain will faithfully hash whatever the server returns,
including a server that reports `accepted` when the safety controller actually
rejected motion. LIMITATIONS.md already draws this line: Phase 1 attests the
gateway boundary and not what happens on the other side, and any claim that
relies on server-side proof is Phase 2 work.

The example's third scenario makes the stakes concrete: cMCP authorizes the
request, the physical state changes, and the independent controller then rejects
motion on a stale state token. Today the audit chain can prove the request was
authorized under a specific policy. It cannot prove, against a hostile or faulty
tool server, what the controller actually decided or did.

## Why this is embodied-specific

We cannot route actuation through cMCP. The functional-safety controller must
remain an independent authority outside the AI governance plane, for the same
certification reasons that make it trustworthy. So the governance plane cannot
observe the actuation directly. It can only bind the controller's own evidence
after the fact. The design question is how to bind an independent safety
authority's signed execution record into a tamper-evident chain without (a)
implying cMCP observed it, or (b) implying functional-safety certification.

## Proposed direction

Add an optional, backward-compatible way to attach independent execution evidence
to the audit entry for a tool call, distinct from `response_payload_hash`:

- Option A, external attestation block on the audit entry. Extend the AuditEntry
  schema with an optional `external_execution_evidence` object:
  `{ issuer, issuer_key_id, signature, evidence_hash, evidence_type, linked_call_id }`.
  The controller (or a thin signer next to it) produces a signed receipt of its
  decision and, where available, the actuation outcome. cMCP records the receipt
[...]

Issue #293, opened 2026-06-11
Populate response_payload_hash in the forwarding path audit entries


imran-siddique:
The AuditEntry schema carries response_payload_hash (SHA-256 of the canonical response), but the proxy's success path appends the final tool_call entry with request_payload_hash only -- the response hash is never populated. Surfaced by the industrial-embodied-ai example (agentrust-io/examples#16), which needs the audit bundle to bind the controller outcome, not just the request.

Fix: hash the post-scan response content (the same bytes the egress check sees) and include it in the step-6 audit append in proxy.call_tool.

Generated with Claude Code

PR #314, opened 2026-06-16
feat(audit): add optional external_execution_evidence on AuditEntry (#301)

carloshvp:
## Summary
Implements Option A for #301 (confirmed direction). Adds an optional,
independently-signed execution receipt bound to an audit entry, kept distinct
from `response_payload_hash`: `response_payload_hash` is what the gateway
forwarded, `external_execution_evidence` is what an independent authority (for
example a safety controller) attested.

This PR lands the data model, schema, verification, and tests. Proxy ingestion
and the example follow next (see Scope and follow-ups).

## What changed
- `audit/chain.py`: optional `external_execution_evidence` field on `AuditEntry`
  and an `append()` keyword. Serialized uniformly via `asdict` (null when
  absent), so receipt-less entries hash exactly as before and existing evidence
  keeps verifying. No change to the canonical body rule.
- `schemas/audit-entry.schema.json`: optional receipt object (`issuer`,
  `issuer_key_id`, `signature`, `evidence_hash`, `evidence_type`,
  `linked_call_id`), not in `required` so entries that predate the field still
  validate.
- `cmcp_verify.verify_audit_bundle`: opt-in receipt verification via an
  `external_evidence_keys` parameter (issuer_key_id to hex Ed25519 public key).
  When supplied, it checks `linked_call_id == call_id` and the issuer signature
  over the canonical receipt (all fields except `signature`). Receipt-less
  entries and callers without keys are unaffected.
- `LIMITATIONS.md`: what the receipt does and does not prove.
- conformance tests (`TestExternalExecutionEvidence301`): absent-verifies and
  keeps the old hashing, populated-verifies, tampered-fails,
  linked_call_id-mismatch-fails, unknown-issuer-key-fails.

## Validation
- `pytest tests/unit tests/conformance`: 701 passed
- `ruff`, `mypy`, `bandit`: clean
- DCO signed off

Every existing audit and verify test passes unchanged, so the change is
backward compatible.

## Scope and follow-ups
Kept out of this PR on purpose, happy to do them next:
1. Proxy ingestion: how a controller-signed receipt reaches the gateway. My
   default is to let the upstream MCP server include the receipt in its tool
   response and have the proxy bind it at the existing post-scan audit append
   (the #293 path). Flagging the transport convention for your input before I
   wire it.
2. The industrial-embodied-ai example third scenario showing a controller-signed
   reject receipt (examples repo).
3. Verification-side coordination is filed as trace-spec#34.

Aside: `audit-entry.schema.json` already drifts from the dataclass (`detail`,
`workflow_id`, and a few `entry_type` / `policy_decision` enum values are
missing). Left out of scope here, happy to reconcile separately.

@imran-siddique tagging you as requested.


imran-siddique (commented):
Thanks for this - attaching controller-signed external execution evidence to individual audit entries is exactly the kind of receipt binding that compliance use cases need. The overall approach makes sense. A few things to work through before this is ready:

**Schema gaps**

`evidence_hash` has no pattern constraint. Every other hash field in the schema uses `"pattern": "^sha(256|384):[0-9a-f]+"`. Add one here so schema validation can reject malformed hashes before they reach the verifier.

`issuer_key_id` has no documented format. Is it a hex-encoded SHA-256 of the public key (matching `agent_manifest.py`'s `_key_id()`)? A DID key identifier? Something else? Pick a format and enforce it in the schema.

**Specification**

What is the `evidence_hash` pre-image exactly? The field is described as a "hash of the evidence" but there is no canonical definition of what bytes are hashed, in what encoding, with what canonicalization. An implementer reading the schema today cannot produce a verifiable `evidence_hash` without guessing. This needs a computation spec in `docs/spec/verification-library.md` (or a new doc).

`evidence_type` is a free-form string. Even an initial enumeration of documented values (e.g., `tee-signed-receipt`, `controller-jwt`, `opaque-receipt`) would help both implementers and verifiers. Free-form types make it impossible to write a strict verifier.

`docs/spec/verification-library.md` is not updated to document the new `external_evidence_keys` parameter in `verify_audit_bundle()`. The spec doc should describe what keys are expected, what the verification flow is, and what `EXTERNAL_EVIDENCE_VERIFICATION_FAILED` would look like as an error code.

**Type inconsistency**

`verify.py` declares `external_evidence_keys: dict[str, str] | None`. The sibling PR #315 uses `trusted_agent_manifest_keys: dict[str, bytes]` (raw public key bytes). If these two verification paths are going to coexist in the same library, the key type should be consistent. String-encoded keys require a documented encoding convention that currently doesn't exist.

**Testing**

The conformance tests cover present/absent/tampered/call_id-mismatch cases - that is a solid starting point. But there is no end-to-end test that runs the full path: `chain.append(..., external_execution_evidence=...)` → export bundle → `verify_audit_bundle(external_evidence_keys=...)`. The conformance suite exercises the schema and mismatch detection, not the cryptographic verification path. Add an integration test that uses a real key pair, produces a valid signature, and confirms it passes.

**TRACE Claim**

When a session's audit chain contains entries with `external_execution_evidence`, can a verifier tell from the TRACE Claim alone that external evidence was bound? Currently the Claim carries `audit_chain_tip` but no evidence-presence flag. Is that intentional? If so, call it out in `LIMITATIONS.md`.

carloshvp:
Addressed the review feedback in `4f82a66`.

What changed:
- Tightened `external_execution_evidence` schema validation:
  - `evidence_hash` now requires a `sha256:`/`sha384:` prefixed hash.
  - `issuer_key_id` is now specified/enforced as lowercase hex SHA-256 of the raw Ed25519 issuer public key.
  - `evidence_type` is restricted to documented initial values.
- Updated `verify_audit_bundle()` so `external_evidence_keys` is `dict[str, bytes]`, matching the raw-key convention used by the Agent Manifest verifier path.
- Added key-id self-checking: `issuer_key_id` must match `sha256(public_key_bytes)`.
- Documented the detached evidence hash pre-image, receipt signing pre-image, verification flow, trusted key map format, and `EXTERNAL_EVIDENCE_VERIFICATION_FAILED` semantics in `verification-library.md` and `error-codes.md`.
- Added the TRACE limitation note: the claim commits the audit chain tip but does not carry a separate external-evidence-present flag; verifiers detect evidence by fetching/verifying the committed audit bundle.
- Added an end-to-end conformance test for `chain.append(..., external_execution_evidence=...)` -> exported signed bundle -> `verify_audit_bundle(..., external_evidence_keys=...)` with a real Ed25519 keypair.
- Softened the receipt-less wording so it no longer implies newly emitted entries omit the null evidence field.

Validation run locally:
- `uv run ruff check src/ tests/`
- `uv run mypy src/cmcp_runtime src/cmcp_verify`
- `uv run pytest tests/unit -q` (`645 passed`)
- `uv run pytest tests/conformance -q` (`58 passed`)

GitHub currently shows the PR mergeable with the visible `gate` check passing.

carloshvp:
Follow-up pushed for the #301 path:

- added runtime ingestion for well-formed `external_execution_evidence` objects returned by upstream JSON tool responses
- kept `response_payload_hash` bound to the exact response bytes returned to the caller
- added proxy unit coverage for binding valid receipts and ignoring malformed receipt-like fields
- added docs for the response-field convention
- opened the embodied-AI examples follow-up as agentrust-io/examples#29, with a live local run verifying `controller receipts: verified (2)`

Validation rerun before push:
- `uv run pytest tests/unit/test_mcp_proxy.py tests/conformance/test_audit_conformance.py -q`
- `uv run ruff check src/ tests/ && uv run mypy src/cmcp_runtime src/cmcp_verify`
- `uv run pytest tests/unit -q`
- `uv run pytest tests/conformance -q`

carloshvp:
@imran-siddique this should be ready for re-review when you have a chance.

The original review feedback was addressed in `4f82a66`, and the follow-up runtime ingestion path landed in `29ef037`. Checks are green, and the trust boundary remains explicit: `response_payload_hash` is what the gateway forwarded; `external_execution_evidence` is what the independent controller attested.

Once this lands, I can re-pin and undraft the embodied-AI example follow-up in agentrust-io/examples#29.

imran-siddique:
Superseded by #330 (rebased onto main after #329 landed, all conflicts resolved). Thanks @carloshvp!

PR #315, opened 2026-06-16
feat(session): bind Agent Manifest identity to Trust Records (#302)

### LINKED ISSUE #302: Bind the Agent Manifest identity to the runtime session subject and TRACE Trust Record
## Context

The industrial-embodied-ai example (agentrust-io/examples#16, refined in #18
and #27) documents this under "Evidence boundaries": the Agent Manifest declares
a signed agent identity and the approved prompt, policy and tool hashes, but the
current preview does not bind that identity to the runtime session. The README
states it plainly: the current preview does not cryptographically bind that
manifest identity to the runtime session.

This is the identity member of the same binding family as #301 (independent
controller evidence) and #126 (tool transcript). #301 binds what the controller
did. #126 binds what tools were called. This issue binds who acted.

## The gap

The committed Trust Record makes it concrete:

- `trace.subject` is `spiffe://cmcp.gateway/session/<uuid>`, described in
  trace-claim.schema.json as "SPIFFE URI identifying the gateway session".
- The Agent Manifest `agent_id` is
  `spiffe://factory.example/agent/material-movement/dev`, signed together with
  the prompt, policy and tool hashes.

Nothing links the two. cMCP loads the policy bundle and catalog by hash, so an
offline check can confirm the loaded hashes match the manifest's bound hashes.
But cMCP never ingests the manifest, never checks that the identity acting in the
session is the manifest's `agent_id`, and the Trust Record carries no agent
identity at all, only the session. Verification therefore proves that some
session ran an approved policy and catalog. It does not prove that the agent the
operator reviewed is the agent that acted.

## Why this matters for embodied agents

Embodied agents restart, get redeployed, and get replaced. Identity persistence,
proof that the agent that acted is exactly the one that was reviewed, down to its
prompt, policy and tools, is the property that survives those restarts. Without
the binding, an operator can run a different agent (a different system prompt, a
different planner, a different identity) under the same approved policy and
catalog, and the evidence chain will not register the substitution. Permission
and policy hashes are pinned; the actor is not. For a system that moves
machinery, who actually drove this session is not a detail.

Both identities are already SPIFFE URIs, so the binding has a natural spine, and
cMCP already issues SVIDs conditioned on TEE attestation (#96).

## Proposed direction

- Option A, bind at the gateway. At session create, cMCP ingests the signed
  Agent Manifest, verifies the issuer signature against a configured trust
  anchor, checks that the session's authenticated subject (the agent SVID from
  #96) matches `manifest.agent_id`, and checks that the loaded policy bundle and
  catalog hashes equal the manifest's bound hashes. The Trust Record then carries
  the `manifest_id` and the bound `agent_id` alongside the session subject, and
  the verifier asserts the binding. Strongest option, and it reuses the SPIFFE
  path already in place.
- Option B, bind at verification only. Keep manifest ingestion out of the
  gateway and have the verifier cross-check the offline manifest against the
  Trust Record: session subject consistent with `manifest.agent_id`, hashes in
  agreement. Lighter, but it only detects a mismatch after the fact and cannot
  refuse a mismatched run at the boundary.
- Option C, status quo, documented as permanent. Rejected: identity persistence
  is a core property of the trust model, and leaving it undefined invites readers
  to treat a signed manifest as proof of who acted, which it is not today.

Recommendation: Option A as the binding spine, with the Option B verifier check
kept as defense in depth so offline evidence is self-checking.

## Scope for a first iteration

- Manifest ingestion at session create, with issuer signature verification
  against a configured trust anchor.
- Subject-to-agent_id binding check (SVID equals `manifest.agent_id`), plus
  policy bundle and catalog hash agreement with the manifest, default deny on
  mismatch.
- Carry `manifest_id` and the bound `agent_id` into the Trust Record, and have
  cmcp-verify assert the binding.
- Conformance tests: a matching manifest binds and verifies; a mismatched
  agent_id, a tampered manifest signature, and a policy or catalog hash drift
  each fail closed.
- Update LIMITATIONS.md and the session and trace-claim spec to state exactly
  what the binding does and does not prove.
- Update the industrial-embodied-ai example to show the bound case. This needs
  the committed signed evidence regenerated from a live run, the same capture
  step flagged in examples#27.
- Coordinate the canonical manifest shape and signed_fields with the
  agent-manifest SDK (agentrust-io/agent-manifest).

## Non-goals

- This binds identity, not behavior. It does not establish that the reviewed
  agent behaved well, only that the reviewed agent is the one that acted.
- It does not, by itself, prove logical-agent continuity across restarts or
  replacements. Binding a session to its reviewed manifest is the prerequisite;
  reputation and continuity across sessions are separate work.
- Not a replacement for #301 (controller execution evidence) or #126 (tool
  transcript). The three bind different facts: who acted, what tools ran, and
  what the controller did.

I am happy to own this end to end (spec text, gateway change, conformance tests,
and the example update), in a personal capacity.

### LINKED ISSUE #317: Use agent-manifest SDK verify_manifest in the gateway instead of the in-repo reimplementation
## Background
cMCP currently verifies Agent Manifests with an in-repo reimplementation of the RFC 8785 signing pre-image (`src/cmcp_runtime/agent_manifest.py`, added in #302 / #315). The Agent Manifest SDK now exports the canonical verification API from its package root (agentrust-io/agent-manifest#176), and documents the runtime binding contract (agentrust-io/agent-manifest#177, closes agent-manifest#175).

## Ask
Update the gateway to call `agent_manifest.verify_manifest()` (and `signing_pre_image()`) from the SDK instead of maintaining a parallel implementation, so the two cannot drift on canonicalization (notably JCS number formatting and the `hitl_record.approvals` normalization).

## Before this lands
- Align the trusted-keys type: the SDK's `VerificationContext.trusted_keys` is `dict[str, str]` (issuer key_id to base64url-encoded Ed25519 public key), while the cMCP path holds raw `bytes`. The gateway must base64url-encode its keys when calling the SDK. See the format note in agent-manifest `docs/api-reference/verification.md`.
- Confirm the SDK is a suitable runtime dependency for the gateway (it is already a transitive part of the stack).

## Outcome
One canonical verifier, shared between signer, SDK, and gateway. Closes the loop opened in #301-era review and agent-manifest#175.

carloshvp:
## Summary
Implements the cMCP runtime/spec/verifier side of #302: a configured signed Agent Manifest can now be bound to the runtime session identity and carried into the TRACE Trust Record as `gateway.agent_identity`.

Closes #302.
Closes #317.

## What changed
- Added an Agent Manifest binding helper that delegates signature, canonical signing pre-image, and runtime artifact verification to the `agent-manifest` SDK root API: `verify_manifest()`, `VerificationContext`, and `signing_pre_image()`.
- Kept cMCP's runtime-facing trust anchor shape as `dict[str, bytes]`, with a boundary adapter that base64url-encodes issuer Ed25519 public keys for the SDK `trusted_keys: dict[str, str]` contract.
- Added optional `agent_manifest` config with `path`, `trust_anchor_path`, and `authenticated_subject`.
- Startup now fails closed when a configured manifest signature, subject, expiry, policy hash, or catalog hash does not match the loaded runtime inputs.
- Session close now emits `gateway.agent_identity` in the Trust Record, while keeping `trace.subject` as the gateway issuing identity and `gateway.session_id` as the session identifier.
- `cmcp_verify.verify_trace_claim` can optionally cross-check a signed manifest and trusted issuer keys against `gateway.agent_identity` for offline defense in depth.
- `cmcp verify` now accepts `--agent-manifest` and `--agent-manifest-trust-anchor`.
- Updated trace schema, config docs, session-policy spec, verifier spec, error registry, and limitations.

## Validation
- `uv run pytest tests/unit/test_agent_manifest.py tests/unit/test_startup.py tests/unit/test_verify.py -q` (46 passed)
- `uv run ruff check src/cmcp_runtime/agent_manifest.py tests/unit/test_agent_manifest.py tests/unit/test_startup.py tests/unit/test_verify.py`
- `uv run mypy src/cmcp_runtime src/cmcp_verify`
- `uv run pytest tests/unit -q` (667 passed)
- `uv run pytest tests/conformance -q` (51 passed)
- Verified the helper against the existing industrial example `agent-manifest.json` and `manifest-public-key.json` from `agentrust-io/examples` during the initial integration pass.

## Scope note
The regenerated industrial-embodied-ai example evidence lives in `agentrust-io/examples`, so this PR leaves that as a companion follow-up rather than committing cross-repo artifacts here.

@imran-siddique tagging you as requested on #302.


imran-siddique (commented):
Thanks for this - this is a substantial piece of work and the direction is right. Binding a signed Agent Manifest to a session at startup, failing closed on any mismatch, and carrying the identity in the TRACE claim is exactly the right threat model for #302. A few things to address before merge:

**Subject source is invisible to verifiers**

`verify_agent_manifest_binding()` has an `allow_dev_subject_from_manifest=True` path that falls back to using `agent_id` as the authenticated subject when no mTLS subject is provided. Nothing in the returned `AgentManifestBinding`, and nothing in the emitted TRACE claim, distinguishes this from a subject that was verified via SVID/mTLS. An auditor reading the TRACE record cannot tell whether `authenticated_subject` was checked against a real transport identity or came from the manifest itself. The TRACE claim's `gateway.agent_identity` should carry a `subject_source` field (e.g., `config` vs `svid`) so verifiers know the strength of the assertion. Without this, dev-mode deployments produce claims that look identical to production claims.

**`expires_at` is never checked**

`verify_agent_manifest_binding()` reads and validates `issued_at` exists, but it never compares `expires_at` to the current time. A manifest past its expiry is accepted without warning. Add an expiry check (or at minimum a warning log and a flag in the binding result).

**`hitl_record.approvals` normalization is undocumented**

`signing_pre_image()` zeros out `hitl_record.approvals` before signing. This is correct for HITL workflows where approvals are added post-signing, but neither `LIMITATIONS.md` nor `session-policy.md` documents this normalization. Auditors who want to verify the signing pre-image need to know exactly which fields are excluded. Add a sentence in `session-policy.md`'s Agent Manifest Identity Binding section.

**Canonicalization test vectors missing**

`_serialize_json` is a hand-rolled RFC 8785 JSON serializer. This is the correctness-critical path - a canonicalization bug here means valid manifests are rejected or, worse, forged manifests are accepted. The test suite currently has no vectors from the RFC 8785 conformance suite (Appendix A of the RFC). Add at least the basic test vectors to `test_agent_manifest.py` so CI will catch any divergence from the reference implementation.

**`session-policy.md` example has wrong `subject`**

The JSON example in the new Agent Manifest Identity Binding section shows:
```json
{ "trace": { "subject": "spiffe://cmcp.gateway/session/<session-id>" } }
```
In TRACE v0.2, `subject` identifies the issuing system or key, not the session. The session identifier lives in `gateway.session_id`. This conflates two distinct identities and will confuse implementers reading the spec. Fix the example to use the correct subject (the TEE gateway identity).

**`CMCP_POLICY_HASH` env var has no implementation**

`docs/configuration.md` adds `CMCP_POLICY_HASH` to the env var table, but there is no code in this PR that implements it. If this is a planned follow-up, it should not land in this doc yet. If it is implemented elsewhere, point to the issue.

**Dead state in `SessionState`**

`session/state.py` gets three new fields (`agent_manifest_id`, `agent_id`, `authenticated_subject`) that are set in `create_session()` but never read in `close_session()` - which pulls the binding directly from `ctx.agent_manifest`. Either these fields are dead state (remove them), or they serve a future purpose (add a comment explaining what). As written they create the impression that session state tracks the manifest identity, but the close path ignores them.

**Missing test for dev bypass path**

There is no test that exercises `allow_dev_subject_from_manifest=True`. Given that this is a security-relevant bypass that changes the subject provenance, it warrants an explicit test confirming the behavior and a follow-up item to add the `subject_source` signal described above.

carloshvp:
Addressed the review feedback in `2e92b4e`.

What changed:
- Added `gateway.agent_identity.subject_source` so verifiers can distinguish configured subject input from dev-mode manifest fallback, and wired it through the schema/model/verifier.
- Added fail-closed `expires_at` validation for Agent Manifests.
- Documented `hitl_record.approvals` normalization in the Agent Manifest signing pre-image.
- Added RFC 8785/JCS canonicalization vectors, including UTF-16 property ordering, and fixed the canonicalizer to preserve source Unicode code points.
- Fixed the `session-policy.md` example so `trace.subject` identifies the gateway issuer and `gateway.session_id` carries the session identifier; runtime claim generation now matches that shape.
- Removed the unused Agent Manifest fields from `SessionState`.
- Added explicit test coverage for the dev-mode subject fallback path.
- Clarified the `CMCP_POLICY_HASH` docs to point at the existing startup integrity check.

Validation run locally:
- `uv run ruff check src/ tests/`
- `uv run mypy src/cmcp_runtime src/cmcp_verify`
- `uv run pytest tests/unit -q` (`667 passed`)
- `uv run pytest tests/conformance -q` (`51 passed`)

GitHub currently shows the PR mergeable with the visible `gate` check passing.

imran-siddique:
Holding this until the in-repo RFC 8785 reimplementation is replaced with the `agent-manifest` SDK call, per issue #317. The path forward:

1. agent-manifest#176 merges (exports `verify_manifest` from package root — ready to merge now)
2. Update this PR to call `verify_manifest()` instead of the local reimplementation
3. Then this is ready

The reimplementation is not wrong, but we've already decided in #317 that it should be delegated to the SDK. Merging as-is locks in debt we'd have to unwind immediately after.

carloshvp:
Addressed the Agent Manifest SDK delegation blocker in `daaa0c1`.

What changed:
- Replaced cMCP's local Agent Manifest RFC 8785/signature verification path with calls to the `agent-manifest` SDK root API: `verify_manifest()`, `VerificationContext`, and `signing_pre_image()`.
- Kept cMCP's runtime-facing trust-anchor shape as `dict[str, bytes]`, and added a narrow adapter that base64url-encodes Ed25519 public keys for the SDK's `trusted_keys: dict[str, str]` contract.
- Kept the session-specific binding checks around the SDK result: subject provenance, authenticated SPIFFE subject equality, manifest expiry handling, and fail-closed mapping for SDK mismatch/expired/unverifiable results.
- Added `agent-manifest` as a runtime dependency pinned to the merged SDK commit that exports `verify_manifest` from the package root. This can become a normal version constraint after the next SDK release is cut.
- Updated docs to say cMCP delegates Agent Manifest verification to the SDK and documents the raw-bytes-to-base64url trust-key adapter.
- Replaced local canonicalization-vector tests with SDK delegation/adapter tests, since RFC 8785 coverage now belongs to the SDK.

Validation run locally:
- `uv run pytest tests/unit/test_agent_manifest.py tests/unit/test_startup.py tests/unit/test_verify.py -q` (`46 passed`)
- `uv run ruff check src/cmcp_runtime/agent_manifest.py tests/unit/test_agent_manifest.py tests/unit/test_startup.py tests/unit/test_verify.py`
- `uv run mypy src/cmcp_runtime src/cmcp_verify`
- `uv run pytest tests/unit -q` (`667 passed`)
- `uv run pytest tests/conformance -q` (`51 passed`)

GitHub shows the PR updated to head `daaa0c1`, with the visible `gate` check passing.

imran-siddique:
Fork deleted; rebased onto main and re-opening as PR from agentrust-io/cmcp directly. All three commits preserved.

Issue #126, opened 2026-06-06
[phase-2] tool_transcript binding — hash-chain MCP call log into Trust Record


imran-siddique:
## What and why

The TRACE Trust Record `tool_transcript` field should cryptographically bind the full tool call log to the Trust Record. In Phase 1, this field contains only the aggregate call summary (`call_count`, `transcript_uri`). In Phase 2, it should contain a hash of the full transcript that is verifiable offline.

The GTC Berlin demo requires showing a `tool_transcript` where a regulator can see:
```json
"tool_transcript": [
  {"tool": "document_reader", "data_class": "confidential", "decision": "allow"},
  {"tool": "credit_score_lookup", "data_class": "confidential", "decision": "allow"},
  {"tool": "risk_report_writer", "data_class": "internal", "decision": "allow"}
]
```

And verify that this list was signed by the TEE key and matches the audit chain root.

## Design
- The audit chain `chain_tip` hash becomes the `tool_transcript.hash` value in the Trust Record
- The full transcript (all `AuditEntry` objects) is available via `/audit/export` (issue #55)
- `tool_transcript` in the TRACE Claim: `{hash: "sha256:...", call_count: N, entries: [{tool, data_class, decision}]}`
- The `entries` array uses the audit chain entries, not raw tool call parameters (no PII leakage)

## Acceptance criteria
- [ ] `GatewayClaim.trace.tool_transcript` includes `hash = audit_chain.chain_tip`
- [ ] Each entry in `tool_transcript.entries` has: `tool_name`, `data_class` (from catalog), `decision` (allow/deny/advisory_deny)
- [ ] `tool_transcript.hash` is verifiable: SHA-256 of canonical JSON of the entries array
- [ ] Integration test: run 3 tool calls → verify `tool_transcript.call_count == 3` and entries match
- [ ] No raw tool arguments or response data in `tool_transcript` (privacy-preserving)

## Notes
Depends on Phase 2 TEE policy enforcement (issue above) — the entries are most meaningful when they're TEE-signed. Target: Q3 2026.

imran-siddique:
[...]

Issue #124, opened 2026-06-06
[phase-2] Policy enforcement inside TEE — Cedar evaluation runs in the enclave, not at API boundary


imran-siddique:
## What and why

Phase 1 enforces Cedar policy at the API layer — the gateway process evaluates policy before forwarding calls to upstream MCP servers. In Phase 1, a compromised host OS can still bypass this (it can intercept the gateway process).

Phase 2 moves the Cedar evaluation inside the TEE boundary so policy decisions are hardware-protected. The Cedar engine must run inside the enclave; the policy bundle must be measured into the TEE attestation report; the enforcement result must be signed with the TEE key.

## Architecture change

Phase 1 (current):
  Agent → [cMCP API layer → Cedar eval → forward] → Upstream MCP server

Phase 2 (target):
  Agent → [cMCP API layer] → TEE enclave: [Cedar eval + session state + audit chain] → [forward or deny] → Upstream MCP server

The TEE boundary is at the Cedar evaluation step. The enclave holds:
- The Cedar policy bundle (loaded and measured at startup)
- The session state (sensitivity tags, call history)
- The audit chain (hash-chained, TEE-signed)

## Acceptance criteria
- [ ] Cedar evaluation runs inside the enclave — the policy bundle is measured into the TEE attestation report
- [ ] `policy.bundle_hash` in the TRACE Trust Record is the hash measured at enclave load time (tamper-evident)
- [ ] `policy.enforcement_mode = "enforce"` means a deny inside the enclave cannot be bypassed by the host
- [ ] All Phase 1 tests continue to pass
- [ ] Conformance tests POLICY-TEE-001 through POLICY-TEE-005 added and passing
- [ ] TRACE Trust Record `policy.bundle_hash` matches the value measured at startup

## Notes
This is the critical path item for the GTC Berlin demo. Without it, the demo shows software-only governance (useful, but not the hardware-proof story). Target: end of Q3 2026.

arian-gogani:
moving Cedar evaluation inside the TEE closes the operator-trust gap for policy decisions. the receipt layer composes cleanly: the Cedar evaluation produces a ALLOW/DENY verdict inside the enclave, and the enclave-attested signature over the verdict is the evidence that the policy ran in a trusted environment.

[...]

PR #316, opened 2026-06-17
feat(audit): privacy-preserving tool_transcript entries (#126)

aryanputta:
Partially addresses #126.

## What
Adds a per-call `entries` array to the cMCP TRACE Claim `trace.tool_transcript`, so a regulator can read the tool-call decision trail offline:

```json
"tool_transcript": {
  "hash": "sha256:<audit-chain tip>",
  "call_count": 3,
  "entries": [
    {"tool_name": "document_reader", "data_class": "confidential", "decision": "allow"},
    {"tool_name": "credit_score_lookup", "data_class": "confidential", "decision": "allow"},
    {"tool_name": "risk_report_writer", "data_class": "internal", "decision": "advisory_deny"}
  ]
}
```

Each entry is built from the audit chain, not from raw tool-call parameters or response bodies, so the transcript stays privacy-preserving (no PII).

## Acceptance criteria
- [x] `tool_transcript.hash == audit_chain.chain_tip` (criterion 1, unchanged binding).
- [x] Each entry has `tool_name`, `data_class` (from the catalog `sensitivity_level`), `decision` (allow/deny/advisory_deny/redact/fault/n/a).
- [x] No raw tool arguments or response data in the transcript (privacy-preserving) — covered by `test_transcript_entries_carry_no_payloads`.
- [x] Tests: 3 tool calls produce 3 entries; `call_count == 3`; entries match; hash roundtrip.
- [ ] **TEE-signing of the entries (criterion's privacy/integrity goal) lands with the Phase-2 in-enclave path (#124).** Today the whole RuntimeClaim is Ed25519-signed at the gateway boundary; moving the entries derivation inside the enclave is the #124 follow-up.

## Design notes / decisions for review
- **Hash semantics.** The issue Design section says `tool_transcript.hash = audit_chain.chain_tip`; one acceptance bullet instead reads "SHA-256 of canonical JSON of the entries". These conflict. I kept `hash = chain_tip` (binds the *full* transcript via the chain) and added a separate `transcript_entries_hash()` so a verifier can recompute the entries digest independently. Happy to fold this into a dedicated field if maintainers prefer the second reading.
- **`data_class` source.** Taken from the catalog entry `sensitivity_level` for the tool, falling back to `"unknown"` when the tool is not in the catalog.
- **Model placement.** `tool_transcript` is defined in the cMCP claim schema (not pulled from `agentrust-trace`), so `entries` is added cleanly without a cross-repo model change. A matching `trace-spec` `tool_transcript.entries` field can follow if the canonical spec wants to carry the same array.

## Tests
`tests/unit/test_trace_claim.py`: 30 pass locally (5 new). Pre-existing failures elsewhere in the suite are unrelated (missing `agent_os`/Cedar backend in a bare checkout; fail identically on main). Ruff clean. DCO signed-off.


imran-siddique (commented):
The overall design is right: deriving entries from the audit chain (not from raw payloads) keeps the transcript privacy-preserving, and keeping `tool_transcript.hash` as the audit-chain tip preserves the existing binding semantics. A few things to address before this can merge.

**1. `audit.db` committed**

The diff includes a binary SQLite file at repo root. This is a test-run artifact and must not be committed. Strip it from the branch and add `*.db` (or at minimum `audit.db`) to `.gitignore`. Committing binary database files bloats history, causes spurious conflicts, and leaks local run state.

**2. `transcript_entries_hash()` is a utility with nothing to verify against**

The PR adds `transcript_entries_hash(entries)` and the docstring says "a verifier recomputes this from the entries and checks it against the digest the profile carries." But the profile does not carry that digest anywhere. `tool_transcript.hash` is the audit-chain tip, not the entries digest. There is no second field in the schema, the model, or the claim output to carry the entries hash.

This needs a decision: either add a `transcript_entries_hash` field to `ToolTranscriptOut` and `trace-claim.schema.json` so a verifier actually has something to compare against, or remove `transcript_entries_hash()` entirely (it currently promises offline verification but provides no anchor for it). Leaving an exported utility function whose docstring describes a verification step that is not implementable will confuse integrators.

You flagged this as an open question in the PR body — it needs to be resolved before merge.

**3. `data_class` is a free-form string; `decision` is properly enumerated**

`decision` has a clean `Literal` enum. `data_class` is a bare `str` that falls back to `"unknown"`. The catalog's `sensitivity_level` values presumably have a bounded set — if so, add a `Literal` or `Enum` constraint here too, or at minimum document the expected values so a verifier knows what to expect. Otherwise a typo in a catalog entry silently passes schema validation.

---

Items 1 and 2 need to be resolved before merge. Item 3 is worth addressing but not a hard blocker if the catalog sensitivity level is intentionally open-ended.

carloshvp:
Adding one small follow-up suggestion to Imran’s review: would it be worth adding a session-level regression test for the actual `SessionManager.close_session` derivation path?

The current tests cover passing `transcript_entries` directly into `generate_trace_claim`, but #126’s runtime behavior depends on deriving entries from audit-chain `tool_call` entries plus catalog metadata. A test could append a few tool calls, close the session with a catalog, and assert order, `call_count == len(entries)`, `data_class` from `sensitivity_level`, and no payload fields in serialized entries.

Small semantics question too: the schema allows `"fault"`, but the derivation appears focused on `entry_type == "tool_call"`. If `fault` or `egress_denied` entries with a tool name are intentionally outside `tool_transcript.entries`, maybe that should be captured in the test/docs so consumers know exactly what `entries` represents.

aryanputta:
Thanks, addressed Imran's blockers and Carlos's follow-up in c15437e.\n\n- Removed the committed audit.db artifact and added *.db to .gitignore so local SQLite state does not reappear.\n- Resolved the unanchored transcript_entries_hash() issue by removing the helper and its test rather than adding a new claim field; tool_transcript.hash remains the audit-chain tip as discussed.\n- Documented data_class as the catalog-defined sensitivity_level copied from the tool catalog, with unknown only when the tool is absent. This is now reflected in the Pydantic model and JSON schema.\n- Added a SessionManager.close_session regression test that derives entries from audit-chain tool_call entries plus catalog metadata, preserves order, checks call_count == len(entries), verifies no payload fields are serialized, and shows fault / egress_denied audit events remain outside tool_transcript.entries.\n\nFocused verification:\n\nPYTHONPATH=/private/tmp/agentrust-cmcp-pr316/src:/private/tmp/trace-spec/src python3 -m pytest tests/unit/test_trace_claim.py tests/unit/test_session_manager.py\n→ 49 passed\n\npython3 -m ruff check src/cmcp_runtime/audit/trace_claim.py src/cmcp_runtime/session/manager.py tests/unit/test_trace_claim.py tests/unit/test_session_manager.py\n→ All checks passed

Issue #371, opened 2026-06-30
security: make cnf.jwk<->report_data binding fatal for SEV-SNP/TDX; fix TDX report_data offset


imran-siddique:
Tracked follow-up from the June 2026 org-wide security hardening review. Wave 1/2 fixes are merged and published; this is remaining hardening.

For SEV-SNP/TDX the cnf.jwk<->report_data binding is currently advisory (non-fatal) and the TDX report_data is read from a best-effort/likely-wrong offset. Extract report_data from the parsed raw_evidence at the correct documented offset and require it to equal jwk_thumbprint(cnf.jwk.x); a mismatch must be fatal for all hardware platforms.

imran-siddique:
Status: the cnf/report_data binding is now fail-closed via PR #386 (SNP). The TDX report_data offset will be pinned by the capture script's known report_data prefix (cmcp-tdx-fixture-v1) once the Azure TDX quote fixture lands; the TDX scaffold PR carries a skipped test that asserts the offset then.

imran-siddique:
Status note on the TDX portion of this issue, after auditing `src/cmcp_verify/tdx.py` on `main`.

**Already done (no action needed):**
- SNP `report_data` binding is now fatal (#390, merged).
- The TDX **DCAP quote** path is correct and fail-closed: `verify_tdx_quote` reads `report_data` at `_TD_BODY_REPORT_DATA_OFF = 520` (the TD-body offset, added in #387 referencing this issue) and sets `failure_reason = "report_data_mismatch"` on a mismatch. This is the real attestation path.

**Residual (the part still open):**
`verify_tdx_measurement` (the raw-TDREPORT branch, ~lines 138-148) still:
1. reads `report_data` from `raw_evidence[0x08:0x08+64]`, and
2. treats a mismatch as **non-fatal** (it only appends `report_data` to `verified_fields` on a match; a mismatch is silently ignored).

Two cautions for whoever picks this up:
[...]

Issue #370, opened 2026-06-30
security: verify attestation report signatures (VCEK / DCAP / EK)


imran-siddique:
Tracked follow-up from the June 2026 org-wide security hardening review. Wave 1/2 fixes are merged and published; this is remaining hardening.

The verifier currently treats the platform report signature as unverified (documented in LIMITATIONS.md). Implement real report-signature verification: SEV-SNP VCEK chain from AMD KDS; Intel TDX Quote via QVL/PCS; TPM AK cert + tpm2_checkquote. Needs vendor libraries + real-hardware test vectors. Until then status must not claim hardware-backed verification.

imran-siddique:
Status: SNP report-signature + VCEK->ASK->ARK chain verification is implemented and fail-closed in PR #386 (tested against a synthetic chain: valid verifies, tampered fails closed, wrong pinned ARK fails closed; one test skipped pending a real Azure SNP fixture). TDX (DCAP) is being scaffolded in a separate draft PR. TPM EK chain still TODO. Real Azure SNP/TDX fixtures come from the capture scripts (Product/Platform/attestation-capture/ in the notes store).

imran-siddique:
### Residual scope (post #386/#387)

SNP VCEK->ASK->ARK chain + signature landed (#386) and offline DCAP TDX quote verification landed (#387). The **TPM AK/EK cert-chain + `tpm2_checkquote` signature path is still TODO** (only `qualifying_data` was fixed in #360), so a TPM quote is currently unauthenticated: a forged TPMS_ATTEST with correct magic and matching `qualifying_data` passes and reports `verified`. TPM analogue of the SNP hole #386 closed.

**Producer (`src/cmcp_runtime/tee/tpm.py`, lines ~106-114):** `ectx.quote(...)` returns `quoted, _signature` but the **signature is discarded** and neither the AK pub nor the EK cert are captured. Evidence container `cmcp_runtime/tee/base.py` `AttestationReport` has only `raw_evidence`. Must capture TPMT_SIGNATURE + AK pub + EK cert (prefer an optional cert-chain field mirroring the SNP passport model; keep old blobs degrading to `unverified`, not error).
[...]

PR #389, opened 2026-07-05
fix(verify): enforce report_data binding and VCEK chain+signature in SEV-SNP verifier (#384)

### LINKED ISSUE #384: Harden SEV-SNP verifier: verify VCEK->ASK->ARK chain and make report_data binding fatal (port approach from ca2a)
## Summary

`cmcp`'s SEV-SNP verifier (`src/cmcp_verify/sev_snp.py`) lags the equivalent in `ca2a` (`src/ca2a_verify/sev_snp.py`). ca2a already verifies the full VCEK -> ASK -> ARK chain to the AMD root, verifies the report signature with the VCEK key, and treats a `report_data` mismatch as fatal. cmcp does none of these. This is the same "verify-the-trust-anchor" class we care about: without chain verification the "measurement matches" check is self-referential, and a non-fatal `report_data` check means the confirmation-key binding / freshness is not enforced.

## Current behavior (cmcp)

`verify_sev_snp_measurement` in `src/cmcp_verify/sev_snp.py`:

1. **`report_data` mismatch is non-fatal.** Docstring: "report_data (nonce) if provided (mismatch is not fatal)." On mismatch the code simply omits `report_data` from `verified_fields` instead of failing. `report_data` carries the confirmation-key binding / freshness nonce, so when a value is expected a mismatch should fail closed.
2. **VCEK/VLEK cert-chain verification is not implemented** (explicitly out of scope; always placed in `unverified_fields`). The report signature is never checked against the AMD root, so the "measurement matches" comparison hashes the report's own `measurement` field and compares it to the claim's own `measurement` string, both attacker-supplied.

## Requested changes

1. When `report_data_hex` is supplied and does not equal the report's `report_data`, set `verified = False` with `PUBLIC_KEY_NOT_BOUND` (binding) or `ATTESTATION_STALE` (freshness), as appropriate.
2. Port ca2a's chain verification: fetch VCEK by `chip_id` + `reported_tcb` from AMD KDS, verify VCEK -> ASK -> ARK against a trusted AMD root, then verify the report signature over the report body with the VCEK key. `ca2a/src/ca2a_verify/sev_snp.py` is the reference implementation.
3. Apply the same treatment to the TDX (Intel PCS/DCAP) and TPM paths.
4. Guardrail: ensure the dispatcher can never return `VERIFIED` while `vcek_cert_chain` (or the TDX/TPM equivalent) is in `unverified_fields` -- only `PARTIALLY_VERIFIED`.

## Notes

- Keep the "software-only scope" language for dev-mode records; this is about not over-claiming on the hardware path.
- Internal security-review finding; suggested triage: security / high for item 1 (quick fix), feature for item 2.

imran-siddique:
Hardens the cmcp SEV-SNP verifier (issue #384), porting the approach already used in `ca2a_verify`.

## Changes
- **`report_data` mismatch is now fatal.** Previously a mismatch was silently ignored (`verified` stayed `True`), so an SNP report for a different enclave whose measurement happened to match would pass. `report_data` carries the confirmation-key binding / freshness nonce.
- **Real VCEK → ASK → ARK chain + report-signature verification.** Adds `verify_cert_chain` and ECDSA-P384/SHA-384 report-signature checks (adapted from `ca2a_verify`). Run when a `vcek_chain` is supplied in evidence and trusted AMD roots are configured.
- **Dispatcher guard.** `verify_trace_claim` no longer marks `hardware_attestation` verified while `vcek_cert_chain` is unverified; such a claim stays `PARTIALLY_VERIFIED`, never `VERIFIED`.

## Security note
`_TRUSTED_AMD_ROOTS` is intentionally empty and must be populated from **verifier configuration** (never from the claim) to enable full rooting. Until configured, SNP claims remain `PARTIALLY_VERIFIED` (fail-closed). Sourcing the trusted root from the claim would recreate the verify-the-trust-anchor bug this fixes.

## Tests
- New: chain+signature success, untrusted-root rejection, tampered-signature rejection (build a real P-384 ARK/ASK/VCEK chain and sign a report), `report_data`-mismatch-is-fatal, and a dispatcher test proving an SNP report with an unverified chain stays `PARTIALLY_VERIFIED`.
- `tests/unit/test_sev_snp_verify.py` + `tests/unit/test_verify.py`: all pass. Full unit suite: **770 passed**. The 2 `test_agt_evidence.py` failures are pre-existing on `origin/main` (unrelated; addressed by the separate `fix/agt-evidence-datetime-utc` branch).

## Follow-up
Apply the same treatment to the TDX (Intel PCS/DCAP) and TPM paths, and wire a production source for the trusted AMD roots.

Closes #384

### DIFF
--- a/src/cmcp_verify/sev_snp.py
+++ b/src/cmcp_verify/sev_snp.py
@@ -1,9 +1,20 @@
-"""AMD SEV-SNP attestation verification -- implements issue #67."""
+"""AMD SEV-SNP attestation verification -- implements issue #67, hardened per #384."""
 from __future__ import annotations
 
 import ctypes
 import hashlib
 from dataclasses import dataclass, field
+
+from cryptography import x509
+from cryptography.exceptions import InvalidSignature
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives.asymmetric.utils import encode_dss_signature
+from cryptography.hazmat.primitives.hashes import SHA256, SHA384
+
+# AMD signs report bytes [0, 0x2A0); the 512-byte signature field follows.
+_SIGNED_BODY_LEN = 0x2A0
+# sig_algo value for ECDSA P-384 with SHA-384 (AMD SEV-SNP ABI).
+_SIG_ALGO_ECDSA_P384_SHA384 = 1
 
 
 class _SnpAttestationReport(ctypes.LittleEndianStructure):
@@ -53,6 +64,10 @@ assert ctypes.sizeof(_SnpAttestationReport) == 0x4A0, (
 _SNP_REPORT_MIN_SIZE = ctypes.sizeof(_SnpAttestationReport)
 
 
+class _SNPChainError(Exception):
+    """Raised when the VCEK certificate chain or report signature fails to verify."""
+
+
 @dataclass
 class SNPVerificationResult:
     verified: bool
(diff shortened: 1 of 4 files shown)


imran-siddique:
Closing as a duplicate. The SNP report-signature + VCEK->ASK->ARK chain verification in this PR was independently implemented and merged as #386 (addressing #370) while this was open, and this branch was built on pre-#386 `main`. The two pieces still missing after #386 — making the `report_data` binding fatal (#371) and never reporting VERIFIED while the VCEK chain is unverified (#370/#372) — are carried forward against current `main` in #390.

PR #426, opened 2026-07-29
fix: refuse to issue trace claim when hardware TEE session report is missing

### LINKED ISSUE #372: harden: fail closed at runtime when per-session attested report cannot be produced
Tracked follow-up from the June 2026 org-wide security hardening review. Wave 1/2 fixes are merged and published; this is remaining hardening. (flagged in #368)

When the per-session TEE report call fails, close_session() falls back to the startup report, which carries no chain-root commitment. The verifier rejects this for hardware platforms (fail-closed at verify), but the runtime still issues the claim. Consider failing closed at runtime (refuse to issue an unbound claim) or clearly marking it.

qubeena07:
## Summary

Closes #372.

When the per session TEE attestation call failed inside create_session, close_session still fell back to the shared startup report and signed a claim off of it. That startup report carries no chain root commitment, so a strict verifier already rejects it, but the runtime handed the claim out anyway as if nothing was wrong.

This change makes close_session raise TeeFault instead of issuing that unbound claim, but only when the platform is a real hardware TEE. In software only dev mode the fallback still succeeds, since that binding was already best effort there and this is not a regression for that path.

## Changes

- src/cmcp_runtime/session/manager.py: close_session now checks whether the per session report is missing and the startup report provider is not software only, and raises TeeFault before building the claim.
- tests/unit/test_session_manager.py: added a test asserting TeeFault is raised on a hardware platform, and clarified the existing dev mode fallback test.

## Test plan

- [x] pytest tests/unit/ passes locally, 837 passed, 7 skipped
- [x] mypy on the changed module clean
- [x] bandit -r src/ clean

### DIFF
--- a/src/cmcp_runtime/session/manager.py
+++ b/src/cmcp_runtime/session/manager.py
@@ -26,7 +26,7 @@ from cmcp_runtime.audit.trace_claim import (
     generate_trace_claim,
 )
 from cmcp_runtime.config import KillSwitchConfig
-from cmcp_runtime.errors import KillSwitchTripped
+from cmcp_runtime.errors import KillSwitchTripped, TeeFault
 from cmcp_runtime.kill_switch import KillSwitchEvaluator
 from cmcp_runtime.session.call_log import CallLog, SessionCallLog
 from cmcp_runtime.session.state import SessionState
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Reviewed the diff. Correct fail-closed fix for #372.

What I checked:
- `TeeFault` exists in `errors.py` (code TEE_FAULT, http_status 500), so the raise is well-typed.
- The `software-only` guard string matches `SoftwareOnlyProvider.provider_name()` exactly, and `AttestationReport.__post_init__` validates provider against `_ALLOWED_PROVIDERS`, so there is no third non-hardware value that would now fail closed by accident.
- Dev-mode behaviour is unchanged and the existing fallback test was updated rather than deleted, with the reasoning in the docstring.
- The new test pins the hardware path and matches on "chain-root commitment".

The behaviour change is that a caller now gets a 500 instead of a claim a strict verifier would reject anyway. That is the right trade: handing back a claim whose report_data carries no chain-root commitment misleads anyone not running the stricter verifier check. The `session_end` entry is already appended before the raise, so the chain still records the close even though no claim is issued, which is what you want for the audit trail.

Only failing check is the maintainer-hold gate.

Issue #436, opened 2026-07-30
security(tee): unverifiable TPM evidence still presents as hardware-attested


imran-siddique:
The SHA-1 fallback path is handled honestly: it sets `measurement_note = "sha1-bank-fallback"`, downgrades the reported provider to `software-only`, and logs that the TRACE claim will not present as hardware-attested.

The unverifiable-evidence path is not handled the same way. When the quote fails, which today is always, the provider still returns `provider="tpm"` with a measurement, and that presents downstream as hardware-attested even though the evidence carries no signature and nothing binds it to a TPM.

**Impact.** cMCP reports a stronger claim than the evidence supports. It is the first thing anyone evaluating the TPM tier will catch, and it undercuts the honest degradation already implemented a few lines above in the same file.

**Fix.** Apply the same downgrade discipline. If no verifiable quote is present, report `software-only` with an explicit note such as `tpm-pcr-read-unsigned`, so the claim tier matches the evidence tier. Revisit once the signature and AK work lands, at which point a signed and chained quote can legitimately present as hardware-attested.

Issue #435, opened 2026-07-30
bug(ci): CODEOWNERS points at src/cmcp_gateway which does not exist, so security-reviewers are never requested


imran-siddique:
`.github/CODEOWNERS` contains:

```
* @agentrust-io/maintainers
src/cmcp_gateway/audit/ @agentrust-io/security-reviewers @agentrust-io/maintainers
src/cmcp_gateway/tee/ @agentrust-io/security-reviewers @agentrust-io/maintainers
src/cmcp_gateway/policy/ @agentrust-io/security-reviewers @agentrust-io/maintainers
```

There is no `src/cmcp_gateway/` directory. The source tree is `src/cmcp_runtime/` and `src/cmcp_verify/`.

**Impact.** All three security-sensitive path rules match nothing, so `@agentrust-io/security-reviewers` is never auto-requested for TEE, audit, or policy changes. They fall through to the catch-all rule instead. Every TPM change under discussion in the current RFC would have merged without a security review request.

**Fix.** Repoint to the real paths and add a CI check that fails when a CODEOWNERS path matches nothing.

Issue #434, opened 2026-07-30
bug(tee): tpm2_pcrread hex parsing uses lstrip and corrupts PCR values with leading zeros


imran-siddique:
```python
# src/cmcp_runtime/tee/tpm.py:203
hex_val = hex_val.strip().lstrip("0x").lstrip("0X")
```

`str.lstrip` strips any leading characters in the argument set, not a prefix. For a PCR value whose hex digits begin with `0`, this strips the leading zeros along with the `0x`.

Consequences:
- An all-zero PCR, common for unused PCRs, becomes the empty string. `bytes.fromhex("")` yields `b""`, so the PCR silently contributes nothing to the concatenated measurement.
- A value such as `0x0abc...` loses its leading zero, becomes odd-length, raises `ValueError`, and hits `continue`, dropping the PCR entirely. The later `len(pcr_values) < 8` check then fails the whole read with a misleading error.

**Fix.** Use `removeprefix("0x")` or slice after an explicit prefix check, and do not silently `continue` when a PCR that was present fails to parse.

Add a unit test feeding `_parse_tpm2_pcrread_output` a block containing PCR values with leading zeros and an all-zero PCR, asserting the parsed bytes round-trip and that eight PCRs are returned.

imran-siddique:
Reproduced, and the impact is larger than described above. The bug does not corrupt one value, it shifts the whole PCR list.

Running the current parser and the fixed one over a realistic `tpm2_pcrread` block containing an all-zero PCR0 and a leading-zero PCR1:

```
expected PCR count: 8
OLD (main)   count=7 all32B=[1, 31, 32] matches_expected=False
    PCR0: got 00...                       expected 000000000000000000000000...
    PCR1: got b1f2e3d4c5a6978877665544...  expected 0af7c1b2c3d4e5f60718293a...
    PCR2: got a1b2c3d4e5f60718293a4b5c...  expected b1f2e3d4c5a6978877665544...
NEW (fixed)  count=8 all32B=[32] matches_expected=True
```

What happens: an all-zero PCR0 strips to the empty string and becomes a single `0x00` byte via the `or "00"` fallback. PCR1 begins `0x0A`, loses its leading zero, becomes odd-length, raises, and is dropped by the `continue`. Every later PCR therefore occupies the wrong index, and only 7 values survive.

[...]

Issue #433, opened 2026-07-30
enhancement(tee): no TCG event log is collected, so PCR digests are uninterpretable


imran-siddique:
The provider ships `sha256(PCR0 || ... || PCR7)` and, when the quote path works, a `TPM2B_ATTEST` blob. No TCG event log is collected and the verifier does not replay one.

**Impact.** A PCR digest without its event log is opaque. A verifier can tell that platform state differs from a known-good value but not what changed, cannot express policy over individual boot components, and cannot debug a mismatch. In practice any allowlist of expected measurements becomes unmaintainable, because every firmware update invalidates the golden value with no way to explain why.

**Fix.** Collect the TCG event log (`/sys/kernel/security/tpm0/binary_bios_measurements` on Linux, Measured Boot APIs on Windows), ship it with the evidence, and have the verifier replay it against the reported PCR values.

Part of the TPM security model RFC.

imran-siddique:
Tracked in the TPM security model RFC: #439.

imran-siddique:
Availability finding from an Azure Trusted Launch vTPM, 2026-07-31.

`/sys/kernel/security/tpm0/binary_bios_measurements` exists on this platform but is **zero bytes**. The Azure Gen2 UEFI does not hand a TCG log to the guest, so there is nothing to collect and event-log replay cannot be exercised on an Azure vTPM at all.

Consequences for this issue:

- The replay code merged in #443 is covered by synthetic logs only. Validating it against a genuine log needs a platform that publishes one, which in practice means physical client hardware, the same tier `#432` is aimed at.
- The collection side must treat an absent or empty log as a normal condition on cloud vTPMs rather than an error, and must not let a missing log downgrade an otherwise valid quote.
- On platforms that do publish a log, a mismatch between replay and reported PCRs must fail, so the two cases have to be distinguished carefully: absent is not the same as does-not-match.

Recorded in `docs/testing/hardware-validation.md`.

imran-siddique:
Closing this. Event-log replay shipped in #443 and is on main as `src/cmcp_verify/tcg_event_log.py`, with the TPM security model recorded in #439.

The collection half cannot be exercised on anything we can provision: `/sys/kernel/security/tpm0/binary_bios_measurements` is 0 bytes on Azure Trusted Launch (measured 2026-07-31), because the Gen2 UEFI hands the guest no TCG log. There is no further code to write against that. Worth reopening if a platform ever surfaces a real log.

PR #439, opened 2026-07-30
docs(spec): TPM security model for non-confidential-compute devices

imran-siddique:
Adds `docs/spec/tpm-security-model.md`, a written security model for the TPM path on devices without confidential compute.

**Why now.** cMCP claims hardware attestation on the TPM path, and that claim was never written down or checked against the code. It does not hold. Anyone evaluating a deployment on a device without confidential compute asks what integrity guarantees a TPM actually provides, and we had no precise answer. This settles it and gives the docs one consistent vocabulary.

**What it says.** Today the TPM path is an unsigned, self-reported PCR digest. The quote signature is discarded (#429), the quote is called with a capability query instead of a key handle so it never succeeds (#430), there is no AK or EK chain (#431), the gateway is never measured into a PCR (#432), and no event log is shipped (#433). It also records the parser bug that corrupted the measurement itself (#434, fixed in #437) and the claim-tiering proposal (#436).

**One design correction worth reading.** The obvious fix for #432 is to extend an application PCR, but PCR 23 and PCR 16 are resettable from locality 0 per the TCG PC Client Platform TPM Profile. An adversary with local code execution can reset and re-extend, which is exactly the adversary this tier targets. The RFC therefore prefers an NV extend index with a write policy, or sealing across non-resettable SRTM PCRs.

**Decision requested.** Approve the P0 through P5 ordering in section 5 and the three-tier vocabulary in section 7. Section 4.7 also needs an explicit call, since the claim-tiering change contradicts an existing test assertion and ships separately.

Approvals needed from @podcastinator, @AaronRoeF, and @katy-gordon before this leaves draft.

Tracking: #429 #430 #431 #432 #433 #434 #435 #436

### DIFF
--- a/docs/spec/tpm-security-model.md
+++ b/docs/spec/tpm-security-model.md
@@ -1,0 +1,110 @@
+# TPM Security Model on Non-Confidential-Compute Devices
+
+**Status:** Draft, awaiting approval
+**Approvers required:** @podcastinator, @AaronRoeF
+**Tracking issues:** #429, #430, #431, #432, #433, #434, #435, #436
+**Supersedes nothing. Complements** [attestation.md](attestation.md) and [threat-model.md](threat-model.md).
+
+---
+
+## 1. Purpose
+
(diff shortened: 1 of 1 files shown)


pforest (approved):
Approved

PR #465, opened 2026-08-02
fix(otel): wire R8 telemetry export into the running gateway

### LINKED ISSUE #456: OTel export has never run against a real collector
#442 shipped OpenTelemetry export for AARM R8: audit entries mirrored as spans, opt-in via `CMCP_OTEL_ENABLED=1` and `pip install cmcp-runtime[otel]`.

The manual-test box in that pull request was left **unticked** on purpose. OpenTelemetry was not installed in the development environment, so the exporter was exercised only in its inert state. Every automated test covers the sink plumbing, failure isolation and the payload allowlist; none covers a span actually arriving.

Worth one run before this ships in a release, because the untested part is exactly where a type error would hide: `span.set_attribute` accepts a constrained set of value types, and the exporter passes audit-entry fields straight through. An `int | None` or a nested value that the SDK rejects would surface only against a live collector.

To close: run a gateway with `CMCP_OTEL_ENABLED=1` against an OTLP collector, confirm spans arrive with the expected attributes, and confirm a denied call sets span status to error. Then tick the box, or fix what it finds.

zohebk8s:
Wires up the AARM R8 telemetry export that #442 shipped but never connected. Refs #456 — full investigation and the live-collector evidence are in [the issue comment](https://github.com/agentrust-io/cmcp/issues/456).

Deliberately does not close #456, so the issue can be closed by hand once the run is confirmed independently.

## The problem

Running a gateway with `CMCP_OTEL_ENABLED=1` and `pip install cmcp-runtime[otel]`, pointed at a real OTLP collector, produced **zero spans**. Two independent gaps, either one sufficient:

1. **The sink was never attached.** `otel_sink_from_env()` was exported and called nowhere in `src/`. `SessionManager` built its `AuditChain` without `sinks=`, so the only chain a running gateway uses had no observer.
2. **No tracer provider was ever installed.** `trace.get_tracer()` returns a proxy yielding `NonRecordingSpan`s until an SDK provider exists, so nothing read `OTEL_EXPORTER_OTLP_ENDPOINT`. Fixing gap 1 alone still exports nothing.

The exporter's payload logic needed no change — it was correct, just disconnected.

## What changed

1. `SessionManager.__init__` builds the sink once and passes it to every chain it creates.
2. New `configure_tracer_provider()`, called from `cmcp start` before the first session. Installs a `TracerProvider` with a `BatchSpanProcessor` over the OTLP exporter, using the SDK and exporter the `[otel]` extra already declared but never imported, driven by the standard `OTEL_EXPORTER_OTLP_*` variables. An embedder that installed its own provider keeps it.
3. That construction is guarded. The SDK validates `OTEL_EXPORTER_OTLP_*` eagerly and raises `ValueError` on values it cannot parse (`OTEL_EXPORTER_OTLP_TIMEOUT=abc`, `OTEL_EXPORTER_OTLP_COMPRESSION=bogus`). Since this runs at startup, an unguarded call would turn a typo in a telemetry variable into a gateway that will not boot.
4. `OtelAuditExporter.enabled` now reports whether a span emitted now would actually be recorded, rather than whether the OpenTelemetry API imported. It previously returned `True` in exactly the dead state.
5. Tests covering the recording path rather than the inert one.

## Verification

| Check | Result |
| --- | --- |
| Suite with `[otel]` | 1034 passed, 7 skipped |
| Suite on base install (no OpenTelemetry) | 1020 passed, 21 skipped |
| ruff / mypy | clean |
| Live gateway to live collector | allowed call `Unset`; denied call `Error` / `policy_decision=deny`; digests only |
| Shutdown flush (30s batch delay, signalled mid-buffer) | all buffered spans arrived |
| Unreachable collector | tool calls unaffected (0.01–0.06s); shutdown bounded ~8.8s vs ~2.2s healthy, self-terminating |
| Malformed `OTEL_EXPORTER_OTLP_*` values | returns `False` and the gateway starts without export, rather than raising |
| 16 concurrent sessions × 25 entries, one shared exporter | 416/416 spans, no cross-session attribute bleed, no sequence gaps, every chain verifies |

Eight of the new tests fail against `main`, including a direct reproduction of #456: a session created through `SessionManager` emits no span.

Live verification was done with `otel/opentelemetry-collector-contrib:0.115.1` against the `examples/minimal` bundle and the mock upstream.

## Notes for review

- No change to `AuditEntry`, entry hashing, or the TRACE Claim. Export remains a read-only mirror attached as a sink, and existing chains still verify.
- Export remains opt-in and a no-op without `CMCP_OTEL_ENABLED`; the base install is unaffected.
- The design question of whether cMCP owns the `TracerProvider` or defers to `opentelemetry-instrument` was settled by #442's own dependency list: the `[otel]` extra declares `opentelemetry-sdk` and `opentelemetry-exporter-otlp-proto-http`, and nothing imported either. Happy to revisit if the intent was different.
- Reproducing the live run currently needs `COPY README.md .` added to the `Dockerfile` — an unrelated packaging bug on `main`, filed separately and deliberately not fixed here.

### DIFF
--- a/src/cmcp_runtime/cli.py
+++ b/src/cmcp_runtime/cli.py
@@ -84,7 +84,12 @@ def start(config: str, enforcement: str | None) -> None:
     import uvicorn
 
     from cmcp_runtime.config import parse_listen_addr
+    from cmcp_runtime.observability.otel import configure_tracer_provider
     from cmcp_runtime.startup import run_startup
+
+    # AARM R8: install the tracer provider before any session exists, so the
+    # first session's entries are exported too. No-op unless CMCP_OTEL_ENABLED.
+    configure_tracer_provider()
 
     ctx = run_startup(config)
 
--- a/src/cmcp_runtime/observability/otel.py
+++ b/src/cmcp_runtime/observability/otel.py
@@ -48,6 +48,7 @@ _otel_trace: Any = None
 _SpanKind: Any = None
 _Status: Any = None
 _StatusCode: Any = None
+_PLACEHOLDER_PROVIDERS: tuple[type, ...] = ()
 
 try:  # pragma: no cover - import-time branch depends on the environment
     from opentelemetry import trace as _imported_trace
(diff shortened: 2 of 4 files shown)


imran-siddique (approved):
Reviewed the diff and the tests. This closes a real gap: `otel_sink_from_env()` existed but nothing installed an SDK tracer provider and nothing attached the sink to the chain a running gateway uses, so `CMCP_OTEL_ENABLED=1` exported nothing.

What I checked:
- `configure_tracer_provider()` is called in `cli.py start` before `run_startup`, so the first session's entries are covered.
- It no-ops when export is not requested, when the optional dep is missing, and when an embedder already installed a provider. It never replaces an existing one.
- The broad `except Exception` around provider construction is correct here. A malformed `OTEL_EXPORTER_OTLP_*` value must not stop the gateway booting, and the parametrized test pins that.
- `_PLACEHOLDER_PROVIDERS` resolved via getattr degrades `enabled` to optimistic rather than switching export off on an OTel rename. Reasonable tradeoff, and the test documents it.
- Sinks are built once in `SessionManager.__init__` rather than per session, and run after the entry is durable, so a collector outage cannot break the chain.
- `is_recording()` early return in `_export` avoids ~19 wasted `set_attribute` calls per entry on the hot path.
- No payload fields added to `_EXPORTED_FIELDS`, no secrets in the diff.

The shutdown-latency note in the docstring (about 6s extra against a dead collector) is the kind of measured detail that belongs there. Good.

Only failing check is the maintainer-hold gate.

Issue #453, opened 2026-08-01
security(tee): source and pin the Global Virtual TPM CA - 03 root, or Azure provenance is host-dependent


imran-siddique:
Split out of #431, which stays closed: its code deliverable shipped and works. This is the gap that code cannot close on its own.

## The problem

Azure Trusted Launch presents **two** vTPM attestation-key certificate hierarchies concurrently at NV index `0x01C101D0`. Measured on hardware:

| | `D2s_v5`, eastus, 2026-07-31 | `D2s_v7`, eastus2, 2026-08-01 |
|---|---|---|
| Certificate size | 1596 bytes | 994 bytes |
| Issuer | `Azure Cloud Virtual TPM CA - 11` | `Global Virtual TPM CA - 03` |
| AIA extension | present, 4 caIssuers URIs | **absent entirely** |
| Chain to a pinnable root | yes, reaches `Azure Virtual TPM Root Certificate Authority 2023` | **no** |

`cmcp_verify/tpm_roots.py` pins only the 2023 root, which the CA-11 hierarchy reaches over AIA. On a host presenting `Global Virtual TPM CA - 03` there is no AIA to walk, and `tpm2_getcap handles-nv-index` confirmed the intermediates are not stored elsewhere in NV as a fallback. So the chain cannot be assembled at all and `verify_tpm_quote` fails with "AK chain root is not among the supplied trusted TPM roots".

That failure is correct: key provenance genuinely cannot be established on such a host with what we have. But the consequence is that **cMCP's TPM provenance story works on some Azure VMs and not others, with no way for an operator to tell in advance.** Details and the corrected doc are in #452.

## What is needed

1. Obtain the `Global Virtual TPM CA - 03` certificate and its chain up to a self-signed root, from a Microsoft-published source that can be cited. It is not fetchable from the AK certificate itself, since there is no AIA.
2. Pin that root alongside `AZURE_VTPM_ROOT_2023_PEM`, keyed by platform so `trusted_roots_for` can return both for Azure.
3. Confirm on hardware that a host presenting this hierarchy then verifies end to end. That means provisioning until a VM of that generation is drawn, which is not directly selectable; `D2s_v7` in eastus2 produced one on 2026-08-01.
4. Decide what a deployment does when neither root matches: fail closed is the current behaviour and is probably right, but the error should name the unrecognised issuer so an operator can act on it rather than reading a generic chain failure.

## Why not just walk AIA

[...]

Issue #447, opened 2026-07-31
consolidate TPM quote verification onto agent-manifest instead of three divergent copies


imran-siddique:
The same TPM quote verification primitive exists three times across the stack, at three different assurance levels.

| | `agent-manifest` | `cmcp` | `ca2a` |
|---|---|---|---|
| Lines | 205 | ~350 | 98 |
| Structural parse | Yes | Yes | Yes |
| Signature verified | Yes (RSA + ECDSA) | Yes, since #444 | **No** |
| AK chain + pinned root | Yes, SHA-256 fingerprint | **No** (#431) | **No** |
| PCR digest binding | Yes | Partial | **No** |
| Collects a quote | verifier only | Yes | **No PCR read at all** |

`ca2a/src/ca2a_runtime/tee/tpm.py` is parse-only: no `pcr_read`, no quote, no signature check, no AK. So cA2A has no working TPM attestation today.

## This was already designed

The 0.5.0 changelog says the verification API was generalized "so cmcp and ca2a can delegate their full SNP/TDX/TPM crypto to this package (via PyPI) without changing behavior or rewriting their test fixtures."

The pieces are all in place and were never used for TPM:

- `verify_tpm_quote` and `parse_tpm_quote` are in `__all__` of the released `agent-manifest` 0.7.0.
- cmcp already declares `agent-manifest>=0.6.1`; cA2A declares `>=0.5`.

No new agent-manifest release is required.

## Proven against real hardware

`agent_manifest.verify_tpm_quote` was run against a genuine Azure Trusted Launch vTPM quote (`Standard_D2s_v5`, eastus, manufacturer MSFT), with the platform AK certificate from NV `0x01C101D0` and the full chain fetched over AIA:

```
CN=<vm-id>.TrustedVM.Azure.windows.net
  CN=Azure Cloud Virtual TPM CA - 11
    CN=Azure Cloud Virtual TPM CA 2025
      CN=Azure Virtual TPM Root Certificate Authority 2023, O=Microsoft Corporation  (self-signed)
```

Results with that root pinned:

```
full verify        : True
wrong nonce        : False
tampered attest    : False
untrusted root     : rejected (TpmVerificationError)
```

So **#431 needs no new cryptographic code**. Delegating gets signature verification and a pinned chain at once.

## Plan

[...]

Issue #431, opened 2026-07-30
security(tee): no AK or EK provisioning, so ek_cert_chain is unconditionally unverified


imran-siddique:
`src/cmcp_verify/tpm.py` always appends `ek_cert_chain` to `unverified_fields` with `details["ek_cert_chain_validation"] = "ek_cert_chain_validation_requires_ca_lookup"`. There is no AK provisioning in the runtime and no EK certificate is collected or shipped with the evidence.

**Impact.** Even once the signature is retained and checked, nothing ties the signing key to genuine TPM hardware. A software-emulated TPM, or an attacker-generated key, produces evidence indistinguishable from a discrete TPM.

**Fix.**
1. Create a restricted signing AK and certify it against the EK.
2. Ship the EK certificate with the attestation evidence.
3. Verify EK certificate to TPM vendor intermediate to vendor root at verification time.

Open dependency: vendor CA root distribution. This has to cover firmware TPMs as well as discrete ones, since firmware TPMs are what most client hardware ships with, and there is no convenient published path for obtaining and pinning vendor EK CA roots today.

Part of the TPM security model RFC.

imran-siddique:
Tracked in the TPM security model RFC: #439.

imran-siddique:
Hardware findings from an Azure Trusted Launch vTPM on 2026-07-31 change the approach for this issue.

**Credential activation is not the route on Azure.** The only NV index defined on the VM is `0x01C101D0`. There is no EK certificate, at `0x01C00002` or anywhere else. `tpm2_createek` succeeds but yields an uncertified key, so there is nothing to activate a credential against.

**The platform already gives us a certified AK.** Persistent handle `0x81000003` holds exactly the key that the certificate at `0x01C101D0` certifies, confirmed by comparing public keys, and that handle produces a quote directly. The certificate carries EKU `2.23.133.8.3`, the TCG AIK certificate OID.

**The chain is reachable, contrary to what the docs recorded.** The note in `docs/testing/hardware-validation.md` said the certificate carries no AIA extension and its intermediate is not fetchable. Azure has since changed its vTPM PKI. The issuer is now `CN=Azure Cloud Virtual TPM CA - 11`, not `CN=Global Virtual TPM CA - 03`, and there are four AIA CA Issuers URIs. Fetching from the public CDN and verifying gives:

```
CN=<vm-id>.TrustedVM.Azure.windows.net
  signed by CN=Azure Cloud Virtual TPM CA - 11
[...]

PR #469, opened 2026-08-03
fix(verify): authenticate TPM quotes and let evidence reach the verifier (#370)

zohebk8s:
Contributes to #370, the TPM half of the residual scope. Does not close it; see *Not covered*.

## Why

A TPM quote was unauthenticated. The `tpm2` branch of `verify_trace_claim` called only `verify_tpm_measurement`, which takes no signature parameter. `verify_tpm_quote_chained` already existed and was tested, but nothing in the production path called it.

A/B on byte-identical input, only `verify.py` and `tpm.py` differing:

```
forged TPMS_ATTEST: correct magic, matching qualifying_data, no signature, no chain
  before:  hardware_attestation in verified_fields = True
  after :  hardware_attestation in verified_fields = False
```

`VERIFIED` is gated on `hardware_attestation` not being unverified, so this is what enforces the issue's "status must not claim hardware-backed verification". Before this change the forgery was held below `VERIFIED` only by an unrelated failure, not by any check of the evidence.

Evidence also could not reach the verifier. `RuntimeInfo` in agentrust-trace is `extra="forbid"`, so a claim carrying `raw_evidence` / `quote_signature` / `cert_chain` under `trace.runtime` was `CLAIM_MALFORMED` before the platform branch ran.

Both halves land together deliberately. Transport on its own would remove the `CLAIM_MALFORMED` that was incidentally masking the hole, turning a reporting bug into acceptance of a forged attestation.

## What

**Verifier.** New `verify_ak_ek_chain()` in `cmcp_verify/tpm.py`, mirroring `verify_vcek_chain()`. `trusted_tpm_ca_pem` threaded through `verify_trace_claim` alongside `trusted_ark_pem`, at the sites named in the issue. Signature verified with the AK pub taken from the chain leaf. Supplied-but-invalid material is fatal (`HARDWARE_ATTESTATION_FAILED`); absent material degrades to `unverified`, as SNP does.

**Transport.** Signed evidence travels as `gateway.attestation_evidence`. `GatewayAddenda` is cmcp's own model, documented for fields outside the canonical TRACE spec, so no upstream release is needed. `trace.runtime` is still read as a fallback for older claims.

**base64.** The gateway wrote base64url unpadded, the verifier used standard `b64decode`, so the two halves could not have interoperated. The reader now accepts both. Nothing caught this because no test carried evidence through both sides.

A claim built through the claim builder, with evidence populated the way the gateway populates it, now reaches `status: verified` with `failure_reason: None`.

## Decisions worth reviewing

**Back-compatibility is broken in one direction.** Measured both:

```
new verifier + old claim (no evidence)  ->  OK, fields optional, trace.runtime fallback
old verifier + new claim (evidence)     ->  CLAIM_MALFORMED: gateway.attestation_evidence
```

This is not caused by the choice of `gateway`. `GatewayAddenda` and `RuntimeClaim` are both `extra="forbid"`, so any additive field anywhere in the claim is rejected by a verifier built before it, and `verify_trace_claim` never reads `cmcp_version`, so there is no negotiation path. "Evidence travels with the claim" and "old verifiers keep working" cannot both hold under the current models. Options are laid out in the issue comment; this PR implements the first, minus the version bump, which is not mine to decide.

**`AK->EK` is not attempted.** An EK is a restricted decryption key: it cannot sign, cannot issue an AK certificate, and cannot appear in the AK's issuance path. `verify_cert_chain` rejects the shape. Only `EK->manufacturer-CA` is a certificate path, implemented as a separately supplied chain. This matches the #431 pivot to the platform-certified AK at `0x81000003`.

**EK identification needs the EKU and `ca=False`.** The TCG EK EKU `2.23.133.8.1` appears on Azure's issuing CAs (`Azure Cloud Virtual TPM CA - 11`, `... CA 2025`), where it means "may issue EK certificates". Matching on the EKU alone credited `ek_cert_chain` for a chain containing no EK. Regression test included.

**`_cert_signed_by` no longer exists.** #447 replaced it with `agent_manifest.verify_cert_chain`, so this mirrors `verify_vcek_chain()` rather than reviving a deleted helper.

**`verify_tpm_quote_chained` is left uncalled.** The EK chain is a separate path and the links need naming individually, so the wiring uses `verify_ak_ek_chain` plus `verify_quote_signature`. Happy to collapse the two into one entry point if preferred.

## Not covered

- The producer still does not capture an EK certificate.
- SEV-SNP has the same transport defect: `verify_sev_snp_measurement` reads `cert_chain` from `trace.runtime`, which `RuntimeInfo` forbids, so the VCEK verification merged in #386 does not engage for a schema-valid claim either. Not migrated here since it touches merged code. Worth its own issue.
- Manufacturer-root pinning is #453.
- `SessionManager.close_session()` is not driven end to end with real evidence bytes. Both sides of that seam are unit tested; the existing session tests use mocks.

The first three are recorded in `LIMITATIONS.md`.

## Tests

Mirrors `test_snp_signature_verify.py`: synthetic chains plus a signed attest, exercised valid / tampered / wrong-root / missing-chain, plus an end-to-end claim through the producer path.

```
full suite   1041 passed, 22 skipped   (baseline 1020, 22 on main)
added           21 tests, all new; no existing test changed behaviour
ruff         All checks passed
mypy         no issues, 16 source files
bandit       no findings
```

### DIFF
--- a/LIMITATIONS.md
+++ b/LIMITATIONS.md
@@ -54,6 +54,40 @@ The catalog binds each tool name to a specific upstream server identity, which p
 - **Threat classes T1 through T4 are not covered.** These are the rogue administrator, host OS compromise, post-incident audit log reconstruction, and policy substitution threats described in the [threat model](docs/spec/threat-model.md). All four require a hardware TEE to close. In software-only mode, all four remain open.
 - **TRACE Claims are partially verified only.** The `cmcp_verify` library returns `status: partially_verified` and reports `hardware_attestation: software-only mode -- not hardware-backed`. Claims produced in dev mode must not be presented as hardware-attested proof to auditors or regulators.
 
+## What attestation verification establishes, and what it does not
+
+`cmcp_verify` reports `status: verified` only when the platform evidence is
+cryptographically checked; a claim whose report signature or certificate chain
+is unverified stays `partially_verified` and is never presented as
+hardware-backed (issue #370). What that check covers differs by platform:
+
+- **AMD SEV-SNP** — report signature plus the VCEK → ASK → ARK chain, with the
+  ARK pinned by the operator.
+- **Intel TDX** — DCAP quote signature and the PCK chain to a pinned Intel SGX
+  Root CA.
+- **TPM 2.0** — the `TPMT_SIGNATURE` over the `TPMS_ATTEST`, verified with the
+  attestation key, plus the AK certificate chain to a manufacturer CA the
+  operator pins via `trusted_tpm_ca_pem`. Absent signature or chain material
+  degrades to `unverified`; supplied material that fails is fatal.
+
+Two gaps are worth stating plainly for the TPM path:
+
+- **The attestation key is not bound to a specific TPM.** A verified AK chain
+  proves the key was certified under a CA you pinned. It does not prove the key
+  lives in the endorsed TPM. That binding is TPM credential activation
+  (`TPM2_MakeCredential` / `TPM2_ActivateCredential`), a live challenge-response
+  an offline verifier cannot perform, and it is not implemented. A CA that
+  mis-issues, or a platform CA that certifies a software key, is not caught.
+- **Evidence rides in the cmcp envelope, not the TRACE runtime block.**
+  `RuntimeInfo` in `agentrust-trace` is `extra="forbid"`, so a claim carrying
+  `raw_evidence` / `quote_signature` / `cert_chain` under `trace.runtime` is
+  rejected as `CLAIM_MALFORMED` before platform verification runs. Signed
+  evidence therefore travels as `gateway.attestation_evidence`, a cmcp-owned
+  field. The verifier still reads `trace.runtime` as a fallback so older claims
+  keep working, but that path cannot pass schema validation. **The SEV-SNP
+  `cert_chain` path has not been migrated and remains subject to this**, so its
+  VCEK chain verification does not engage for a schema-valid claim.
+
 ## What cMCP does not do
 
 - **cMCP is not a WAF.** It does not inspect HTTP traffic for SQL injection, XSS, or other web application attack patterns. It operates at the MCP tool call layer, not the HTTP layer.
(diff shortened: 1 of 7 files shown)


imran-siddique (approved):
Reviewed the diff and the reasoning. This closes a real hole: the `tpm2` branch called only `verify_tpm_measurement`, which takes no signature, so a forged TPMS_ATTEST with correct magic and matching qualifying_data reached `hardware_attestation` verified. The A/B in the description is the right way to demonstrate that.

What I checked and agree with:

- Fail-closed shape matches the SNP path. Absent material degrades to unverified, supplied-but-invalid is fatal with `HARDWARE_ATTESTATION_FAILED`, and `hardware_attestation` is only credited when both the measurement and the chain check out.
- `ek_cert_chain` requiring the TCG EKU **and** `ca=False` is correct and non-obvious. Azure carries `2.23.133.8.1` on its issuing CAs, so EKU-only matching would have reported `ek_cert_chain` verified for a chain holding no EK. Good that there is a regression test on it.
- Not attempting AK to EK as a certificate path is right. An EK is a restricted decryption key and cannot appear in the AK issuance path. Credential activation is the real binding and is correctly left out of scope.
- Carrying evidence in `gateway.attestation_evidence` rather than `trace.runtime` is the correct call given `RuntimeInfo` is `extra="forbid"` upstream.
- The base64 mismatch catch is a good find. The gateway wrote base64url unpadded while the verifier used standard `b64decode`, so the two halves could never have interoperated.

Two non-blocking notes:

1. `_decode_evidence` returns `None` on undecodable input, so corrupt base64 in `quote_signature` is classified as absent rather than supplied-but-invalid. That takes the non-fatal path, which is inconsistent with the principle stated elsewhere in this PR. It cannot lift a forged claim to `VERIFIED`, since the field still lands in unverified, so it is a reporting nit rather than a hole. Worth distinguishing decode failure from absence in a follow-up.
2. `_b64`'s docstring says a report object that does not carry these fields at all must not break claim construction, but `report.quote_signature` and `report.attestation_key_chain_pem` are plain attribute reads. Both exist with defaults on `AttestationReport`, so real providers are fine; only a hand-rolled test double would raise. Either tighten the docstring or use `getattr`.

Approving the code. Holding the merge for one decision that is not mine: this breaks old verifiers in the forward direction, since a new claim carrying `gateway.attestation_evidence` fails `extra="forbid"` on any verifier built before the field. That needs a version bump and a release note before it ships, which is the piece you correctly flagged as not yours to decide.

PR #474, opened 2026-08-04
feat: add agent_key_thumbprint scaffold to gateway.agent_identity

### LINKED ISSUE #425: Carry the agent's key thumbprint in gateway.agent_identity
Raised by @l33tdawg in [agentrust-io/.github discussion #15](https://github.com/orgs/agentrust-io/discussions/15) while taking the SAGE integration in-consensus. Opening it here so the work is tracked rather than leaving him to file it.

## Problem

A RuntimeClaim's `cnf` is the **gateway's** TEE-sealed key (`_build_cnf(signing_key)`, subject `spiffe://cmcp.gateway/tee/<prefix>`). The agent rides in `gateway.agent_identity`, which carries `manifest_id`, `agent_id`, `authenticated_subject`, `subject_source`, `issuer`, `issuer_key_id`, `policy_bundle_hash`, and `tool_catalog_hash`.

It does **not** carry anything identifying the agent's own signing key.

That blocks a legitimate downstream pattern: a consumer that wants to check, deterministically and offline, that the party which signed some downstream artifact is the agent the claim describes. In SAGE's case every validator wants to evaluate `thumbprint(on_chain_author_pubkey)` against the claim, with no network call.

The two obvious workarounds are both wrong:

- **Export the gateway's sealed key** so the agent can sign with it. This defeats the point. The claim asserts that a signature came from inside a measured runtime; a key that can be handed to a caller no longer supports that assertion.
- **Compare against `cnf`**, which is the gateway key, so it answers a different question than the one being asked.

## Proposal

Add an optional field to `agent_identity`, something like:

```
agent_key_thumbprint: str   # "sha256:<hex>" over the agent's signing public key
```

The gateway already authenticated the agent, so it is the correct party to attest that binding. The thumbprint sits inside a payload signed by the TEE-sealed key, so a verifier who appraises the claim chain gets the agent-key binding for free, and non-exportability is preserved because only a hash leaves the enclave.

Additive and backward compatible: existing consumers ignore it, and claims without it stay valid.

## Open questions

1. **Digest form.** `sha256:<hex>` for consistency with the rest of the claim, or a JWK thumbprint per RFC 7638 for consistency with `cnf.jwk`? RFC 7638 is arguably more correct for a key, and interoperates with anything already handling `cnf`.
2. **Where the gateway learns the key.** Agent Manifest binds the agent's signing identity, so `manifest_id` may already be enough to resolve it. If so, this field is a convenience that saves a lookup rather than new information, which is still worth having for an offline validator but changes how we describe it.
3. **Optional or required at a conformance level.** Leaning optional, since not every deployment has an agent key distinct from the gateway.

## Note

There is a path that works today without this: the agent signs with its manifest-bound key and the verifier checks that key against the signed manifest record. This issue is about making the binding available from the RuntimeClaim alone, which is what an in-consensus check needs.

qubeena07:
## Summary

Closes #425.

SAGE wanted an offline check that a downstream signature came from the agent a TRACE Claim describes. Digging into it turned up that the issue's own premise does not hold on current code: the Agent Manifest binding carries no agent public key anywhere, and subject_source is a static config value today, not a live authenticated credential. There is no key material anywhere in the runtime to hash yet. This was confirmed and worked through together with the issue author in the comment thread, who agreed to the approach below.

Landed as a nullable, additive field instead. agent_key_thumbprint, an RFC 7638 JWK thumbprint rendered as sha256 hex, sits on AgentIdentityInfo and AgentIdentityOut and stays None today since no code path supplies agent key bytes.

The real behavior change is on the verifier side. verify_trace_claim now fails closed with AGENT_KEY_THUMBPRINT_UNBOUND_SUBJECT on any claim that carries the field while subject_source is not live authenticated, so a future producer cannot launder a config supplied identity into what looks like a hardware attested key binding. This check is unconditional, it runs whether or not the caller requests the existing agent manifest cross check.

Populating the field for real needs either an agent_manifest_sdk schema change carrying the agent's public key, or a live mTLS or challenge response credential wired into the binding. Both are out of scope here and stay tracked in the issue thread.

## Changes

- src/cmcp_runtime/audit/trace_claim.py: agent_key_thumbprint added to AgentIdentityInfo and AgentIdentityOut, threaded through generate_trace_claim.
- src/cmcp_runtime/session/manager.py: pointer comment only, no behavior change, the field is not wired in yet.
- src/cmcp_verify/verify.py: new AGENT_KEY_THUMBPRINT_UNBOUND_SUBJECT error and the subject binding check described above.
- docs/spec/session-policy.md, docs/spec/verification-library.md, CHANGELOG.md: document the field and the verifier rule.
- tests/unit/test_trace_claim.py, tests/unit/test_session_manager.py, tests/unit/test_verify.py: schema round trip, current always None invariant, and the three verifier cases (live authenticated passes, config sourced fails closed, absent field unaffected).

## Test plan

- [x] pytest tests/unit/ passes locally, 931 passed, 7 skipped
- [x] bandit -r src/ clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,11 @@ Five changes below the headline TPM fix, each one a case where cMCP reported mor
 - **SNP `report_data` mismatch is now fatal (#371, #390),** since that field carries the confirmation-key binding and the freshness nonce, and the dispatcher no longer reports `hardware_attestation` as verified while the VCEK chain is unverified (#370, #372); such a claim stays `PARTIALLY_VERIFIED`. Same defect class as the TPM headline, on the AMD path, fixed first.
 
 - **TCG event logs are replayed against the reported PCR values (#443)**, so a log that does not reproduce the quoted digests is detected rather than trusted as narrative.
+- **`gateway.agent_identity.agent_key_thumbprint` scaffold (#425).** SAGE (via l33tdawg, agentrust-io/.github discussion #15) wanted an offline check that a downstream signature came from the agent a TRACE Claim describes. Investigating turned up that the issue's own premise did not hold: `AgentManifestBinding` carries no agent public key anywhere, and `subject_source` is a static config value today, not a live-authenticated credential (`svid` is a valid value but nothing produces it). So there was no key material anywhere in the runtime to hash.
+
+  Landed as a nullable, additive field instead: `agent_key_thumbprint` (RFC 7638 JWK thumbprint, `sha256:<hex>`) on `AgentIdentityInfo`/`AgentIdentityOut`, always `None` today since no code path supplies agent key bytes. The real, non-speculative piece is on the verifier side: `cmcp_verify.verify_trace_claim` now fails closed (`AGENT_KEY_THUMBPRINT_UNBOUND_SUBJECT`) on any claim that carries the field while `subject_source` is not live-authenticated, so a future producer cannot launder a config-supplied identity into something that looks like a hardware-attested key binding.
+
+  Populating the field for real needs either an `agent_manifest_sdk` schema change carrying the agent's public key, or a live mTLS/challenge-response credential wired into `AgentManifestBinding` - both explicitly out of scope here, tracked in the issue thread.
 
 ### Fixed
 
--- a/docs/spec/session-policy.md
+++ b/docs/spec/session-policy.md
@@ -181,6 +181,14 @@ This binding answers "who acted" for the session. It does not replace `trace.sub
 
 Offline verifiers SHOULD cross-check `gateway.agent_identity` against the signed manifest and trusted issuer key. This keeps the runtime boundary check and the evidence artifact self-checking.
 
+`gateway.agent_identity` MAY also carry `agent_key_thumbprint`: an RFC 7638 JWK thumbprint of the agent's own
+signing key, rendered as `sha256:<hex>`, distinct from `issuer_key_id` (the key that signed the manifest).
+It is optional and is omitted from every claim the current runtime can produce, since no code path here has
+access to agent key bytes yet. When a producer does populate it, it MUST only do so while `subject_source`
+names a live-authenticated source (currently `svid`) - never `config` or `manifest-dev`, which are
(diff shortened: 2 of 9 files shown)


imran-siddique:
Thanks for the patience on this — and apologies for the wait.

**The CI failure here is not your change.** All seven failing jobs are the security audit tripping on `cryptography 48.0.1` (PYSEC-2026-3552/3553/3554), which is what a plain resolve produced when this branch was opened because agt-core 4.1.0 declares `cryptography>=46.0.7,<49.0`.

`main` has since gained a "Force cryptography past agt-core's stale cap" step for exactly that. A rebase onto current `main` should clear all of it without touching a line of your code. Happy to do the rebase myself if you'd rather.

On the substance, and this is the part I want to flag as read rather than skimmed: you found that the issue's premise did not hold, said so, worked it through with the issue author, and landed a nullable scaffold instead of manufacturing key material to satisfy the original ask. That is the right call and the harder one to make.

The verifier-side check is the part carrying real weight — failing closed with `AGENT_KEY_THUMBPRINT_UNBOUND_SUBJECT` when the field is present and `subject_source` is not live-authenticated, unconditionally rather than only when the caller asks. That closes the laundering path a future producer would otherwise have: a config-supplied identity dressed as a key binding. Worth keeping that check unconditional even if it later looks redundant.

Review of the diff itself is Imran's; this note is so a five-day-old red build is not read as a verdict on the work.

imran-siddique:
Rebased this onto current main for you, as offered. The only conflict was CHANGELOG.md, where main and your branch both added entries under Unreleased; both are kept. No code changes, and your commit and authorship are intact.

The branch now picks up main's "Force cryptography past agt-core's stale cap" step, which is what the red build was about. CI is running on 402e6cf.

imran-siddique:
Rebase confirmed — nine checks green. The only red is `gate`, which is the maintainer-approval requirement rather than anything about your code, and it clears on review.

Thanks for turning that around quickly. Review is with Imran.

imran-siddique (approved):

qubeena07:
Thanks @imran-siddique, appreciate you catching the CI issue and doing the rebase, and the note on the verifier check landing as intended. Glad the scaffold approach worked out.

Issue #472, opened 2026-08-04
We are pinned to an agt-core generation upstream has already removed (CedarBackend, v4 policy language)


imran-siddique:
## The problem

`PolicyEvaluator` wraps AGT's v4 Cedar backend:

```python
# src/cmcp_runtime/policy/evaluator.py
from agent_os.policies.backends import CedarBackend
```

That module no longer exists upstream. [microsoft/agent-governance-toolkit#3444](https://github.com/microsoft/agent-governance-toolkit/pull/3444) replaced the v4 policy language with ACS v5 across the Python runtime and deleted `agent_os/policies/backends.py`. On AGT `main` there is no `CedarBackend` outside a migration CLI helper.

We do not notice because `agent-governance-toolkit-core>=4.0` resolves to **4.1.0**, tagged 2026-06-09, which still ships the module. Every release after it on that line has moved on.

## Why this matters now

It is already blocking security work. #471 needs `cryptography>=49.0` to close two reachable advisories, and the agt-core releases that permit it are the same ones that removed the API we import. The short-term answer there is a 4.1.1 backport, which is the right call for a CVE fix but does not change the underlying position: we are one dependency resolution away from a broken gateway, and pip will make that choice for us the moment our floor moves or 4.1.0 is yanked.

The exposure is not marginal either. Cedar evaluation is the enforcement path. If it breaks, cMCP allows or denies nothing correctly, and the demos, the web console, and every conformance claim go with it.

## What needs deciding

1. **Migrate to ACS v5.** Understand what replaces `CedarBackend`, whether policy bundle hashing and the `mode="auto"` cedarpy/cli/builtin selection have equivalents, and what happens to the annotation extraction in `_advice_for_deny` that reaches into `_build_cedar_request`. That last one already depends on a private attribute, so it will not survive untouched.
2. **Or vendor the evaluation we need.** We use a small surface: `evaluate(context)`, `_build_cedar_request`, and the backend-mode selection. `cedarpy` is already a direct dependency of ours. Owning ~200 lines against `cedarpy` directly would end the coupling, at the cost of maintaining it.
3. **Or pin exactly and accept it.** `agent-governance-toolkit-core==4.1.0` makes the current state explicit instead of accidental, and buys time. It also means never taking an upstream fix again, including security fixes to agt-core itself.

Option 3 is worth doing immediately regardless of which of 1 or 2 wins, because `>=4.0` currently lets pip pick a release that would break us silently.

## First step

[...]

Issue #471, opened 2026-08-04
Fix two reachable cryptography CVEs: needs an agt-core 4.1.1 with the cap lift backported


imran-siddique:
## Corrected

My first pass on this issue said agt-core 5.0.0's `agentrust-trace<0.3.0` pin looked like a packaging oversight and that lifting it would unblock us. That was wrong on both counts, and the real picture is better in one way and worse in another.

## What is actually true

`cryptography` 48.0.1 carries three advisories. Two are reachable here, through the untrusted `cert_chain` a claim carries into `verify_ak_ek_chain` and `verify_vcek_chain`:

| advisory | what it is | fixed in | reachable |
|---|---|---|---|
| CVE-2026-69249 | duplicate self-signed intermediates cause exponential path building | 49.0.0 | **yes** |
| CVE-2026-69248 | X.509 verifier accepts wildcard DNS names, escaping `permittedSubtrees` | 49.0.0 | **yes** |
| CVE-2026-69247 | Bleichenbacher oracle in PKCS#7 EnvelopedData decryption | 50.0.0 | no, we only parse cert bundles |

**49.0.0 is enough**, and I verified it end to end in a clean venv with agt-core 4.1.0 and `cryptography` forced to 49.0.0:

```
cmcp test suite            1041 passed, 22 skipped
CedarBackend evaluate()    works, policy decisions unchanged
pip-audit (as CI runs it)  No known vulnerabilities found, 1 ignored   -> exit 0
```

So the cap is not load-bearing for anything this package exercises. Take 49.0.0 and CI goes green with the single unreachable advisory already suppressed.

## The one thing blocking it

`agent-governance-toolkit-core` 4.1.0, which is the release pip resolves for us, pins `cryptography>=46.0.7,<49.0`.

The cap was already lifted upstream, in [#3070](https://github.com/microsoft/agent-governance-toolkit/pull/3070) (`e75972f7`, 2026-06-17, a one-line change to `<50.0`). It missed the 4.x line by 8 days: v4.1.0 was tagged 2026-06-09, and the monorepo moved to 5.0.0 on 2026-06-25. So the fix exists in history and has never shipped in a release we can install.

## Why we cannot just take agt-core 5.x

[...]

PR #477, opened 2026-08-06
fix(inspection): stage 3 content classification was dead when AGT is installed

imran-siddique:
## What this fixes

`_classify_sensitivity` called `_agt_redactor.find_credentials(response_text)`. That method **has never existed** on agt-core's `CredentialRedactor`. I checked the pinned 4.1.0 wheel and every commit in AGT history; the real API is `find_matches` (secrets) and `find_pii_matches` (PII spans).

```python
# before
if _AGT_AVAILABLE and _agt_redactor is not None:
    try:
        matches = _agt_redactor.find_credentials(response_text)   # AttributeError, always
        ...
    except Exception:  # nosec B110
        pass                                                      # swallowed
else:
    for pattern, tag in _PII_PATTERNS:                            # never reached with AGT present
        ...
```

The `AttributeError` went into a bare `except Exception: pass`, and the local pattern sweep sat in the `else` branch, so it did not run either. **Whenever agent-os was importable, source 3 of stage 3 contributed nothing.** No response received a content-derived sensitivity tag; classification silently fell back to catalog annotations alone.

The existing tests all patch `_AGT_AVAILABLE` to `False`, so they exercised the fallback and stayed green throughout.

## Changes

- Call `find_matches` and `find_pii_matches`, resolved via `getattr` so a future upstream rename degrades to a logged warning instead of a silent no-op.
- Run the local `_PII_PATTERNS` as a **second pass** rather than an either/or.
- Replace the blanket `except Exception: pass` with a narrowed, logging handler. Swallowing everything is what hid this.

## Why the second pass, and not just the corrected method name

[microsoft/agent-governance-toolkit#3494](https://github.com/microsoft/agent-governance-toolkit/issues/3494): AGT's redactor does not redact a secret **at all** when a suffix is glued to it, so `AKIAIOSFODNN7EXAMPLE_old` passes through whole. That issue's fix PR was closed unmerged on 2026-08-05 and there is no release pending, so relying on AGT alone leaves a live key untagged.

## Tests

Three regression tests, each verified failing on the previous implementation and passing here:

| Test | Scenario |
|---|---|
| `test_local_patterns_still_run_when_agt_lacks_the_method` | AGT object without the expected method |
| `test_local_patterns_still_run_when_agt_raises` | AGT method raising |
| `test_suffixed_secret_still_tagged_despite_agt_3494` | the `_old` suffix case |

`tests/unit/test_stage3_sensitivity.py` and `tests/unit/test_inspection.py`: 56 passed. `ruff` clean, `mypy` clean.

## Not affected

cmcp is **not** exposed to [#3496](https://github.com/microsoft/agent-governance-toolkit/issues/3496). That defect is in `MCPResponseScanner.sanitize_response`, reached only under `ResponsePolicy.SANITIZE`. `proxy.py` leaves the default `BLOCK`, and the pipeline calls `scan_response`.

## Note on the local test environment

`pytest tests/unit` shows 21 collection errors on this branch and on pristine `main` alike. They come from a dev checkout of AGT shadowing the pinned agt-core 4.1.0, so `agent_os.policies.backends` and `mcp_gateway.GovernancePolicy` are absent. That is #472 reproduced locally, and worth noting there: on 5.x, `proxy.py` loses `GovernancePolicy` as well as the evaluator losing `CedarBackend`.

Refs #471, #472.

### DIFF
--- a/src/cmcp_runtime/inspection/pipeline.py
+++ b/src/cmcp_runtime/inspection/pipeline.py
@@ -222,8 +222,16 @@ def _stage4_injection_detection(
                     injection_score=score,
                 )
             return StageResult(stage="injection", decision="allow")
-        except Exception:  # nosec B110
-            pass  # Fall through to regex
+        except Exception:
+            # Falling through to the local patterns is the right behaviour, but
+            # doing it silently means a broken AGT detector looks identical to a
+            # working one while the weaker starter set is what actually ran.
+            _log.warning(
+                "AGT PromptInjectionDetector.detect() failed; "
+                "falling back to local patterns v%s",
+                _PATTERNS_VERSION,
+                exc_info=True,
+            )
 
     # Fallback: regex patterns
     patterns = custom_patterns or _COMPILED_PATTERNS
@@ -423,13 +431,35 @@ class InspectionPipeline:
         self._agt_redactor: Any = None
         self._agt_response_scanner: Any = None
         if _AGT_AVAILABLE:
+            # Constructed independently. Sharing one try block meant a failure in
+            # the first component skipped the other two, so a single upstream API
+            # change disabled three security components at once, silently.
             try:
                 _cfg = DetectionConfig(sensitivity=injection_sensitivity)
                 self._agt_injection_detector = PromptInjectionDetector(config=_cfg)
+            except Exception:
+                _log.warning(
+                    "AGT PromptInjectionDetector unavailable; "
+                    "stage 4 falls back to local patterns v%s",
+                    _PATTERNS_VERSION,
+                    exc_info=True,
+                )
+            try:
                 self._agt_redactor = CredentialRedactor()
+            except Exception:
+                _log.warning(
+                    "AGT CredentialRedactor unavailable; "
+                    "stage 3 falls back to local patterns",
+                    exc_info=True,
+                )
+            try:
                 self._agt_response_scanner = AGTResponseScanner()
-            except Exception:  # nosec B110
-                pass
+            except Exception:
+                _log.warning(
+                    "AGT MCPResponseScanner unavailable; "
+                    "stage 4 loses MCP-specific threat detection",
+                    exc_info=True,
+                )
 
     def run(
         self,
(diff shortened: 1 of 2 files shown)


imran-siddique:
## Second commit: the handlers that hid the first bug

Audited every AGT symbol and method cmcp calls against the pinned agt-core 4.1.0 wheel, since a phantom method surviving in `main` suggested the pattern might repeat.

**Good news on the API surface.** `find_credentials` was the only one that does not exist. Everything else resolves against 4.1.0:

| Call site | Symbol / method | 4.1.0 |
|---|---|---|
| `catalog/scanner.py` | `MCPSecurityScanner.{register_tool, scan_tool, check_rug_pull}` | ok |
| `inspection/pipeline.py` | `PromptInjectionDetector.detect`, `DetectionConfig` | ok |
| `inspection/pipeline.py` | `MCPResponseScanner.scan_response` | ok |
| `mcp/proxy.py` | `MCPGateway.{intercept_tool_call, intercept_tool_response}` | ok |
| `mcp/proxy.py` | `GovernancePolicy` | ok, re-export |
| `mcp/server.py` | `StatelessKernel` | ok |
| `policy/evaluator.py` | `CedarBackend` | ok on 4.x |

On `GovernancePolicy`: it is defined in `agent_os.integrations.base`, and `mcp_gateway.py` imports it at module level, so `from agent_os.mcp_gateway import GovernancePolicy` works at runtime. The `# type: ignore[attr-defined]` is legitimate, not papering over a break.

**What the audit did find** is the error handling around those calls, all in `pipeline.py`:

1. **One try block built all three components.** A failure constructing the first left the other two unbuilt and all three silently `None`. A single upstream API change would disable three security components at once with no log line. Now constructed independently, each logging on failure.
2. **`detect()` failing fell through to local patterns silently.** The fallthrough is correct; the silence is not. A broken detector looked identical to a working one while the weaker starter set was what actually ran. Now logged with the pattern-set version that took over.
3. **`scan_response()` failing was swallowed entirely.** It correctly does not deny, since an errored scanner produced no verdict and stage 4's own detection still runs, but losing MCP-specific coverage for a response should not be invisible. Now logged with the tool name.

`catalog/scanner.py` already did this properly and is unchanged. It logs, sets an explicit `_available = False`, and distinguishes not-installed from failed. That is the pattern the other two now follow.

Two more tests, both verified failing on the previous commit. 58 pass across `test_stage3_sensitivity.py` and `test_inspection.py`; ruff and mypy clean.

## One thing deliberately not changed

When AGT's detector returns a clean verdict, stage 4 returns `allow` immediately and the local pattern set never runs. Only the *exception* path falls back, despite the docstring reading as though the regex set is a general fallback.

Running both and denying if either denies would be more conservative and consistent with the stage 3 fix in the first commit. But it changes deny behaviour and risks false positives in a gateway, and `patterns_version` is reported in every `InspectionResult` as the set that made the decision, which is already misleading when AGT short-circuits. That is a design call, so I have left it for review rather than folding it into a robustness fix.

Issue #479, opened 2026-08-06
Sensitivity vocabulary is fixed, so a deployment cannot express its own classification scheme


imran-siddique:
## What I hit

`SENSITIVITY_ORDER` in `cmcp_runtime/session/state.py` is a hard-coded six-value table:

```
public(0)  pii(1)  confidential(2)  hipaa_phi(3)  mnpi(3)  trade_secret(3)
```

and the catalog schema validates `sensitivity_level` against exactly that set. Correct, and it fails closed — see #478.

The problem is that a regulated public-sector deployment does not classify data that way. National and sector schemes commonly use a four-tier ladder along the lines of Open / Confidential / Secret / Top Secret, and those labels are the ones written in the regulation, in the customer's policy documents, and in the audit report a regulator reads. Today none of the top two can be expressed.

## Why the workaround is unsatisfying

Two options exist and both are bad:

1. **Map onto the built-in vocabulary** — call the top tier `trade_secret` because it happens to rank 3. The transcript then shows a class the customer does not use and the regulation does not name. In an audit that is worse than useless.
2. **Carry the real class as a call argument** and let policy discriminate on it, leaving `sensitivity_level` at a single catalogued value. This works for enforcement, but `ToolTranscriptEntry.data_class` is populated from the catalog entry (`session/manager.py`, ~line 276), so every row in the signed transcript reports the same class regardless of what the call actually touched. The record under-describes what was enforced.

Option 2 is what I ended up doing. The enforcement is right and the evidence is weaker than the enforcement, which is the wrong way round for a product whose whole proposition is that the record is trustworthy.

## What would fix it

Roughly in order of appetite:

- **Configurable ordering.** Let a deployment supply its own `sensitivity_level` vocabulary and ordering in config; the schema validates against that rather than a literal. Keeps the fail-closed property from #478, since the set is still closed, just not hard-coded.
- **Per-call classification.** Allow a call to declare a class in the `_cmcp` metadata block, record it on the audit entry, and prefer it over the catalog default in the transcript. To stay safe it must only ever *raise* the effective class above the catalogued floor, never lower it — a caller must not be able to declare its own data down a tier.

The two compose: configurable vocabulary decides what labels are legal, per-call classification decides which one this call used.

## Why it matters beyond one deployment

[...]

PR #489, opened 2026-08-09
feat: let a deployment add its own sensitivity labels alongside the built in six

qubeena07:
## Summary

Part of #479, the configurable vocabulary piece. Per call classification, the second piece of that issue, is left for a follow up.

The sensitivity vocabulary was six hardcoded labels, public, pii, confidential, hipaa_phi, mnpi, trade_secret, in SENSITIVITY_ORDER, and the catalog schema validated sensitivity_level against exactly that literal enum. A deployment whose own regulation names a tier above trade_secret, an Open, Confidential, Secret, Top Secret ladder for example, had no way to catalogue it truthfully.

sensitivity.vocabulary in config now lets a deployment add new labels at any rank.

## Design decision worth flagging

The vocabulary is additive only, not a full replacement. Response inspection's content pattern detectors emit the built in tags pii and hipaa_phi directly when they spot an SSN, an email, a diagnosis code and so on. If a deployment's configured vocabulary could drop either of those names, those detections would rank at 0 by the ordinary unknown tag default, the same fail open hole #478 closed for the catalog schema, just moved to a different door.

So a vocabulary key that names a built in label is rejected at config load, and effective_sensitivity_order merges the built in table in last regardless of what config parsing did, so a built in name can never be shadowed even by a mistake elsewhere in the chain.

SessionManager and PolicyEvaluator each derive the effective vocabulary once from the same Config, so a session's max_sensitivity string and the sensitivity_level integer Cedar evaluates can never disagree about what a custom label ranks as.

## Changes

- src/cmcp_runtime/config.py: new SensitivityConfig, sensitivity.vocabulary parsing, additive only guarantee enforced at parse time.
- src/cmcp_runtime/session/state.py: effective_sensitivity_order(), SessionState.sensitivity_order field, _max_sensitivity takes an optional order parameter.
- src/cmcp_runtime/session/manager.py, src/cmcp_runtime/policy/evaluator.py: both derive the effective vocabulary once from the same Config and thread it through.
- schemas/catalog-entry.schema.json, src/cmcp_runtime/catalog/loader.py: sensitivity_level enum moved out of the static JSON schema into a Python level check at load time against the effective set, so it stays closed, just not hardcoded. Same error message shape as before, existing tests unchanged.
- src/cmcp_runtime/startup.py: wires config.sensitivity.vocabulary into load_catalog.
- docs/spec/session-policy.md, docs/tutorials/tool-catalog-authoring.md, CHANGELOG.md.

## Test plan

- [x] pytest tests/unit/ passes locally, 996 passed, 8 skipped
- [x] bandit -r src/ clean
- [x] manual check: custom vocabulary loads, collision with a built in label rejected, catalog entry with a custom label rejected without the extension and accepted with it, session ranks a custom label correctly, built in tags like hipaa_phi still rank correctly with a custom vocabulary configured

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Configurable sensitivity vocabulary (#479).** The six built in sensitivity labels (`public`, `pii`, `confidential`, `hipaa_phi`, `mnpi`, `trade_secret`) were the only ones a catalog entry could ever declare, so a regulated deployment whose own scheme names a tier above `trade_secret`, an Open/Confidential/Secret/Top Secret ladder for example, had no way to catalogue it truthfully.
+
+  `sensitivity.vocabulary` in config now lets a deployment add new labels at any rank. It is additive only and enforced at two layers: config parsing rejects a vocabulary key that names a built in label outright, and `session/state.py`'s `effective_sensitivity_order()` merges the built in table in last regardless, so a built in name can never be shadowed even by a bug elsewhere in the chain. That guarantee is not cosmetic: response inspection's content pattern detectors emit the built in tags `pii` and `hipaa_phi` directly when they spot an SSN, an email, a diagnosis code and so on, and if either name had silently dropped out of the effective vocabulary those detections would have ranked at 0, the same fail open hole #478 closed for the catalog schema.
+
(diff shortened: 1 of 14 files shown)


imran-siddique (approved):
Approving. Reviewed in a separate worktree and ran the three test files this touches: 93 passed.

Three things I checked rather than took on trust, all of which you got right.

`effective_sensitivity_order` merges the built ins last, so a colliding config key cannot shadow `pii` or `hipaa_phi` even if the parser somehow let one through. That matters more than it looks: response inspection emits those tags directly, and `order.get(tag, 0)` fails open to rank 0, so a shadowed built in would silently downgrade to public. Defending it in two places for a case that should already be impossible is the right instinct here.

There are only two `SessionState` construction sites. The manager passes the effective order; the other is the benchmark harness, which uses `public`. So no production path can silently fall back to the built in ranking.

Moving `sensitivity_level` out of the JSON schema does not open a hole, because the load time check is still closed, just against a config dependent set, and the schema says why the enum is gone.

One thing for the follow up rather than this PR: `compliance_domain` keeps its hardcoded enum, so a deployment can now name its own sensitivity tier but not its own compliance domain. Whoever hits that will hit it immediately after adopting this.

PR #515, opened 2026-08-15
fix: widen the AIA chain depth cap for TPM attestation keys and log truncation

### LINKED ISSUE #514: AIA chain walk in TPMProvider._chain_from_leaf has zero headroom past the one known Azure depth
Filed separately at imran-siddique's request in #453, where zohebk8s found this while working the CA-03 root sourcing issue.

What I hit

_AIA_MAX_DEPTH is 4 in src/cmcp_runtime/tee/tpm.py line 41, and the walk in TPMProvider._chain_from_leaf (line 333) is

while len(chain) < _AIA_MAX_DEPTH:
    current = chain[-1]
    if current.subject == current.issuer:
        break
    ...
    chain.append(issuer)

The condition is checked before each append, not after. The only real chain measured so far, leaf, Azure Cloud Virtual TPM CA - 11, Azure Cloud Virtual TPM CA 2025, root, is exactly 4 certificates, so the loop happens to reach the self signed root on its third append, then exits on the fourth check. There is no margin left over.

Why this is fragile rather than theoretical

The same investigation in #453 found a second Azure hierarchy, Global Virtual TPM CA - 03, that reaches the same root in only 3 certificates. Two different depths already exist on real fleet hardware today. If a future Azure hierarchy needs one more intermediate than the CA-11 path, walk through it by hand: at len(chain) 4 the while condition is false and the loop exits before ever fetching the root, since the self signed check on the actual root certificate is never reached. The resulting chain is one certificate short and silently missing its root.

What that produces downstream

agent_manifest._cert_chain checks chain[-1].fingerprint() against the trusted root set and raises AK chain root is not among the supplied trusted TPM roots when it does not match. A truncated chain and a chain that legitimately reaches an unpinned root produce the exact same error today, so an operator cannot tell "we did not walk far enough" from "this hierarchy is not one we trust" from the message alone.

What is needed

Raise _AIA_MAX_DEPTH with real margin above the two known depths rather than matching one of them exactly, and add a test that walks a synthetic chain longer than either known hierarchy to prove the cap does not truncate it. Worth also telling the two failure modes apart in the log at the point the walk exits, even if the operator facing error text lives in agent_manifest, a different repo, per zohebk8s's note in #453.

No hardware needed to fix or test this, the existing tests in tests/unit/test_tpm_platform_ak.py already build synthetic certificates for TPMProvider._chain_from_leaf, mocking a longer synthetic AIA walk follows the same pattern.

Related: #453, #449, #452

qubeena07:
## Summary

Closes #514, split out of #453 at imran-siddique's request after zohebk8s found this while sourcing the Azure CA-03 root.

_AIA_MAX_DEPTH in tee/tpm.py bounds how many certificates TPMProvider._chain_from_leaf will walk via the AIA extension when building the chain shipped alongside a TPM attestation key. It was 4, exactly the length of the one real Azure hierarchy measured so far, leaf, Azure Cloud Virtual TPM CA - 11, Azure Cloud Virtual TPM CA 2025, root. The #453 investigation already found a second real hierarchy, Global Virtual TPM CA - 03, that reaches the same root in only 3 certificates, so two different depths already exist on real fleet hardware today. A cap sized to match one of them exactly left zero room for a future or differently configured hierarchy that needs one more hop, in which case the walk stops one certificate short of the root and the shipped chain silently lacks it, then fails downstream verification with the same error a chain that legitimately reaches an untrusted root would produce.

Not a loop logic bug. The while len(chain) < N walk correctly builds any chain up to N total certificates, including one where the root arrives on the Nth certificate. The bug is purely that N was sized to match one observed case exactly rather than with real margin.

## This is a security critical path

src/cmcp_runtime/tee/ is listed in CONTRIBUTING.md as requiring two maintainer approvals and an explicit security impact note, so flagging that plainly. This PR widens a depth cap and adds a diagnostic log. It does not change what is trusted, how a certificate signature is checked, or the fail closed behavior when a chain cannot be built or does not reach a pinned root.

## Changes

- src/cmcp_runtime/tee/tpm.py: _AIA_MAX_DEPTH raised from 4 to 6, comment records both known real depths and that the margin is a judgment call pending more fleet data. A warning is now logged when the walk stops at the depth cap without reaching a self signed root, distinguishing that case from legitimately reaching an untrusted root, in the logs, before it becomes a confusing chain verification failure.
- tests/unit/test_tpm_platform_ak.py: no existing test walked a multi hop AIA chain at all. Added a synthetic chain builder plus two tests, one proving a 5 certificate chain, one hop deeper than either known real depth, completes under the new cap, one proving a chain that does exceed the cap is detected as truncated and logs the new warning.
- CHANGELOG.md.

## Test plan

- [x] pytest tests/unit/test_tpm_platform_ak.py passes locally, 12 passed
- [x] pytest tests/unit/ passes locally, 1084 passed, 8 skipped, plus one pre-existing failure unrelated to this change (test_release_distribution_smoke.py, fails identically on main with this branch's changes stashed, a local dev venv packaging gap, not a regression)
- [x] bandit -r src/ clean
- [x] manually confirmed the constant is 6 and the two new tests exercise both the successful-deeper-chain case and the truncation-and-log case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   The catalog schema conditions its requirements on transport rather than relaxing them, so a network transport still requires `url` and `tls_fingerprint` while stdio requires `spawn`.
 
+### Fixed
+
+- **The AIA chain walk for a TPM attestation key had no margin above the one depth it was measured against (#514).** `TPMProvider._chain_from_leaf` stops walking the certificate chain at `_AIA_MAX_DEPTH`, and that constant was `4`, exactly the length of the one real Azure hierarchy measured so far: leaf, Azure Cloud Virtual TPM CA - 11, Azure Cloud Virtual TPM CA 2025, root. Split out of #453, whose own investigation found a second real hierarchy, `Global Virtual TPM CA - 03`, that reaches the same root in only 3 certificates. Two different depths already exist on real fleet hardware, so a cap sized to match one of them exactly leaves zero room for a future or differently configured hierarchy that needs one more hop: the walk would stop one certificate short of the root, and the resulting chain would ship without it.
+
+  Not a loop logic bug: the walk correctly builds any chain up to the cap, including one where the root arrives on the last allowed certificate. The cap is now `6`, two hops above the deeper of the two known depths rather than matching either exactly, and the walk logs a warning when it stops without reaching a self signed root, so a future truncation is visible in the logs immediately rather than only surfacing later as a chain verification failure that looks identical to a chain that legitimately reaches an untrusted root.
+
 ## [0.4.0] - 2026-08-08
 
 **Anyone running 0.3.0 should upgrade.** On 0.3.0 a forged TPM quote was reported as hardware-attested: the `tpm2` branch of `verify_trace_claim` called only `verify_tpm_measurement`, which takes no signature parameter, so a `TPMS_ATTEST` with the correct magic and matching `qualifying_data` passed with no signature and no certificate chain (#370). The authenticated path existed and was tested; nothing in production called it. This release makes the quote signature and the AK certificate chain the gate on `hardware_attestation`, so evidence that does not chain to a pinned root can no longer report as verified.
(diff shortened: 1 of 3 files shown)


imran-siddique (approved):
Reviewed the AIA traversal boundary, truncation detection, logging behavior, and synthetic multi-hop coverage. Raising the total-certificate cap from 4 to 6 adds bounded headroom without changing trust anchors, certificate validation, or fail-closed downstream verification. The five-certificate and over-cap cases cover the changed behavior, and CI/security checks are green. Approved as the first of the two maintainer approvals required for this security-sensitive path.

Issue #58, opened 2026-06-05
[catalog] Tool catalog drift detection — rug-pull via tool-definition mutation


imran-siddique:
## What and why
Detect when an upstream MCP server changes its tool definitions after the catalog was sealed (P4.2 rug-pull via notifications/tools/list_changed).

## Spec reference
docs/spec/attestation.md §5 (delta detection), docs/spec/threat-model.md (P4.2 rug-pull)

## Acceptance criteria
- [ ] Gateway handles `notifications/tools/list_changed` from upstream servers
- [ ] On receiving the notification, fetches updated tool definitions and computes new definition_hash
- [ ] Compares new definition_hash against catalog entry's approved definition_hash
- [ ] If mismatch: default behavior = CATALOG_DRIFT_DETECTED error for all subsequent calls to this tool; sets tool_catalog.drift_detected = true in subsequent TRACE Claims
- [ ] Configurable alert mode: allow but log drift event in audit chain
- [ ] Unit test: catalog loaded → drift notification received → hash computed → mismatch detected

imran-siddique:
Closing the loop on this, since it was closed on 2026-06-05 with none of its acceptance criteria met against upstream drift. #521 has the full account of what was and was not there.

Upstream drift detection landed in #523 (88d93ec). Against the criteria here:

- [x] On first contact with a server, fetch its advertised tool definitions
- [x] Compare against the approved definition
- [x] On mismatch: `catalog_drift` deny for that session, and `tool_catalog.drift_detected` in subsequent TRACE Claims (that field was hardcoded `False`, so it could never previously be true)
- [x] Configurable alert mode that allows but logs, as `catalog.drift_policy: warn_only`
- [x] Unit tests, in `tests/unit/test_upstream_catalog_drift.py`
- [ ] **Gateway handles `notifications/tools/list_changed`** — deliberately not done

The last one is a design decision rather than an omission, and it is worth stating plainly so it does not get read as done later. Only stdio upstreams could carry that notification today, since HTTP upstreams here are plain request/response. Building it that way would have covered one transport and silently not the other, which is the same partial-coverage shape that produced #521. First contact per server covers both identically, at the cost of not catching mid-session mutation.

`LIMITATIONS.md` states that gap, along with two others: no notification subscription, and a server that will not answer `tools/list` is recorded as unchecked rather than denied.

PR #523, opened 2026-08-17
feat(catalog): detect upstream tool-definition drift, and stop the scanner failing open

### LINKED ISSUE #521: docs: SPEC.md rates P4.2 rug-pull Strong, but the upstream drift detector is not wired into the gateway
## What

`docs/SPEC.md:248` records our coverage of P4.2 as:

> | P4 Supply chain | P4.2 Rug-pull via tool-definition mutation | Strong: catalog hash pinned; runtime drift produces measurement mismatch |

P4.2 is defined at `docs/SPEC.md:67` as an approved upstream server using `notifications/tools/list_changed` to silently mutate a tool definition after the security review concluded, where the tool name and the JSON schema stay identical and only the description changes.

The gateway does not handle that notification. `list_changed` does not appear anywhere under `src/`.

## What is actually there

Three things, none of which is upstream drift detection:

1. `catalog/loader.py:255` verifies `definition_hash` at load time against the catalog file's own contents. That is self-consistency of the catalog, not a comparison against a live server.
2. `mcp/proxy.py:492` compares `self._catalog.catalog_hash` against the hash captured at session start and fails closed with `catalog_drift` on mismatch. This is real and it works, but it detects our own catalog changing under a running gateway. It never talks to the upstream server.
3. `catalog/scanner.py:141`, `CatalogScanner.check_drift`, whose docstring says it is "Called when notifications/tools/list_changed is received from an upstream server (P4.2 rug-pull detection)". It is referenced only by its own module, `tests/unit/test_catalog_scanner.py`, and `experiments/claim3-rug-pull-detection/`. Nothing in `src/` calls it.

## The defence that does exist, and is written down nowhere

`mcp/server.py:469` serves `entry.approved_definition.description` in the `tools/list` response, not whatever the upstream currently advertises. A mutated description therefore does not reach the model through the gateway at all.

That is a genuine structural mitigation and it is arguably a better answer than detection, since it removes the channel rather than watching it. It may well be why #58 was closed. It is not what SPEC.md claims, and it is not stated in `LIMITATIONS.md` or the spec either.

## Related history

#58 set the acceptance criteria for exactly this. First bullet: "Gateway handles `notifications/tools/list_changed` from upstream servers". Closed 2026-06-05 with no closing comment. #348 and #353 cover the experiment and its tests, and both exercise `CatalogScanner` directly rather than through the gateway, so a green suite there says nothing about whether the gateway is protected.

## Second, smaller defect in the same file

When `agent-os-kernel` is not installed, `catalog/scanner.py:85` returns `CatalogScanResult(safe=True, tools_scanned=len(catalog.entries), tools_flagged=0, threats=[])`, and `check_drift` returns `drifted=False` under the same condition.

A control that is absent should report unavailable, not safe. As written, a deployment missing the optional dependency receives a clean scan result it did not earn, and the only signal is an INFO log line at construction.

## Suggested resolution

Two independent decisions:

1. **The claim.** Either downgrade `SPEC.md:248` to describe the structural mitigation accurately, or wire the detector and keep the rating. Either way `LIMITATIONS.md` should carry the honest version, since that file is where this project's claims are supposed to bottom out.
2. **The dead code.** If the structural mitigation is the intended answer, `CatalogScanner` should be removed or explicitly marked as unwired. A tested module sitting next to an experiment directory named `claim3-rug-pull-detection` reads as a shipped control.

## Why now

The GhostSplice disclosure (cross-channel trust fragmentation via malicious MCP servers) uses the tool description as one of its three fragment channels, and it circulated widely on 2026-08-17. The description channel is about to get attention, so it is worth having our own claim about it be exactly right.

imran-siddique:
Closes #521. Meets the substance of #58, which was closed in June with its acceptance criteria unmet.

## What was wrong

`SPEC.md` rates P4.2 (rug-pull via upstream tool-definition mutation) as covered. The detector written for it, `CatalogScanner.check_drift`, was referenced only by its own module, its unit test, and an experiment directory named `claim3-rug-pull-detection`. Nothing in `src/` called it. The `catalog_drift` deny path that does exist compares our own catalog file against itself and never talks to the upstream server.

## What this does

On first contact with each upstream server, the gateway fetches `tools/list` and compares every advertised definition against its approved entry, using a digest of the semantic triple: description, input schema, output schema. Mismatch fails the call closed by default, writes a `catalog_drift` audit entry naming the tool and the kind of change (`definition_changed` or `withdrawn`), and surfaces in the session TRACE Claim.

`catalog.drift_policy: warn_only` routes the call and still records the drift, so an operator can measure their own fleet before enforcing. It is not meant as a resting state.

## Three design points worth reviewing

**The enforcing comparison uses only the standard library.** `CatalogScanner` is backed by an optional dependency. A control that stops working when a dependency is missing is not a control, so the scanner classifies the kind of change and never decides. It is now wired at startup so tool fingerprints get registered, which is what makes classification possible at all.

**The trigger is first contact per server, not `notifications/tools/list_changed`.** Only stdio upstreams could carry that notification today, since HTTP upstreams here are plain request/response. A notification-driven check would have silently covered one transport and not the other, which is the same shape of partial coverage that produced #521. First-contact checking covers both identically. This is the one place where I have not met #58's literal first bullet, and it is deliberate.

**Absence is no longer reported as safety.** `CatalogScanResult` and `DriftResult` gain `available`. Without `agent-os-kernel` they now return `safe=False, tools_scanned=0, available=False` instead of `safe=True, tools_scanned=len(catalog)`. The test that asserted the old behaviour has been rewritten, and its name, `test_scan_catalog_safe_without_agt`, was itself the defect.

Also fixes `ToolCatalogInfo.drift_detected`, hardcoded `False` at `session/manager.py:256`, so the TRACE Claim carried a drift field that could never be true.

## What it does not do

Stated in `LIMITATIONS.md` rather than left implied: no mid-session detection, no notification subscription, and a server that will not answer `tools/list` is recorded as unchecked rather than denied. The gateway already serves the approved description rather than the live one, so a mutated description does not reach the model even in those windows, but that is a structural property of proxying an approved catalog and not a detection result.

`SPEC.md` is untouched, per the decision to keep the rating and put the honest detail in `LIMITATIONS.md`.

## Testing

1209 passing, 23 skipped. 10 new tests in `tests/unit/test_upstream_catalog_drift.py` covering the digest primitive, camelCase and snake_case agreement, description-only mutation, fail-closed and warn-only, withdrawn tools, the unchecked path, once-per-server caching, and drift being caught with the optional scanner absent.

`tests/conformance/test_gateway_conformance.py::test_mcp_initialize` fails on `main` as well, an unrelated `protocolVersion` mismatch (`2025-11-25` against an expected `2026-07-28`). Deselected locally, not touched here.


imran-siddique:
Correcting the last line of the description, which was wrong in a way that matters.

I wrote that `tests/conformance/test_gateway_conformance.py::test_mcp_initialize` is "an unrelated `protocolVersion` mismatch". It is unrelated to this PR, and that part stands. But it is not merely a stale assertion.

CI runs `pytest tests/unit/` only (`.github/workflows/ci.yml:86`), while `pyproject.toml` sets `testpaths = ["tests"]`. So `tests/conformance/` has never been executed by CI, and it has been red on `main` with nothing to say so. Same shape as the defect this PR fixes: a check that exists, that it is reasonable to assume is running, and is not.

Filed as #524, with the counts. 58 conformance tests and 1 integration test are unrun.

The gateway is correct on the failing assertion. Per #509 a legacy `initialize` is answered with the legacy version, and the test encodes the pre-#509 behaviour. Not touched here.

PR #529, opened 2026-08-18
docs(testing): the TDX row covers the quote, not the TDREPORT (#370)

imran-siddique:
Last piece of #370. Part of the CoSAI WS4 Phase 1 review (ws4 #149) close-out.

`docs/testing/hardware-validation.md` exists to enforce one rule: no document describes cMCP as hardware-attested for a platform until a genuine quote from that platform has been verified end to end, and the run is recorded. The TDX row was at risk of being read wider than the run behind it.

The recorded run verifies a **DCAP v4 quote**. `verify_tdx_measurement()` also parses the 1024-byte **TDREPORT_STRUCT** from the `TDX_CMD_GET_REPORT0` ioctl — a different artifact — and no real TDREPORT has been checked against it.

That matters now because #371 found both TDREPORT offsets wrong (`MRTD` read from inside `REPORTMACSTRUCT.report_data`, `REPORTDATA` from the leading RESERVED block) and #527 corrected them against the published Intel ABI. The correction is asserted against the ABI and against a property that needs no hardware to state — a measurement must not move when only the nonce moves. Neither is a capture, and synthetic self-consistency is precisely what this page says is not validation.

Nothing about those offsets being wrong for that long would have been visible to a reader of the table as it stood, which is the reason to write the caveat down rather than carry it in a PR description nobody re-reads.

### DIFF
--- a/docs/testing/hardware-validation.md
+++ b/docs/testing/hardware-validation.md
@@ -13,7 +13,7 @@ been verified end to end by `cmcp_verify`, and the run is recorded below.
 | Platform | Report parsing | Certificate chain | Report signature | Verified against real hardware evidence |
 |---|---|---|---|---|
 | AMD SEV-SNP (Azure CVM, vTPM-rooted) | Yes | Yes, to the real AMD ARK-Milan root | Yes | **Yes**, 2026-07-27, both from a stored capture and **live inside a running CVM** |
-| Intel TDX (GCP C3, non-paravisor) | Yes | Yes, to the pinned Intel SGX Root CA | Yes | **Yes**, 2026-07-27, capture of 2026-07-21 |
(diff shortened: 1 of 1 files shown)


imran-siddique:
@Mayur021 — all required checks are green. Could you take the maintainer approval pass when you have a moment?

Mayur021:
Read the diff. The narrowing is right. The validated run covers the DCAP quote path and no TDREPORT capture exists yet, so the row should not read wider than that. Nothing I would change in the wording.

On the approval itself I would rather not be the one to click it. Phase 2 says these stay in review until an independent maintainer approves, and I am one of the reviewers that record cites, with an issue in the same evidence set. An approval from me would not carry the independence that sentence is claiming. Better from someone outside the review.

pforest (approved):

PR #536, opened 2026-08-20
fix(config): make catalog drift policy configurable

noah-ing:
## What

Allow the existing `catalog.drift_policy` setting through strict configuration validation, reject unknown keys inside the `catalog` block, and document both supported policies.

## Why

This is a narrow follow-up to #523 and #58. The parser already consumed `catalog.drift_policy` and the proxy already enforced it, but `_KNOWN_TOP_KEYS` omitted `catalog`. Every explicit `catalog:` block was therefore rejected before the parser ran; the default `fail_closed` behavior worked only because it required no block.

This reconnects the existing configuration contract without changing drift detection, comparison, or the default policy.

## Security impact

None. This does not touch the TEE boundary, signing path, audit chain, capability tokens, or trust-score inputs. `fail_closed` remains the default, `warn_only` remains an explicit operator choice, and unknown nested keys are rejected so misspellings cannot silently fall back.

## Test plan

- [x] `pytest` passes — 1,179 passed, 6 skipped
- [x] `ruff check` passes
- [x] `mypy` passes — 63 source files
- [x] Manual test performed
- [x] `bandit -r src/ -c pyproject.toml` passes
- [x] `mkdocs build --strict` passes

Manual test: on the parent commit, loading YAML with `catalog.drift_policy: warn_only` raises `ConfigError: Unknown config key 'catalog'`. With this change, it loads as `DriftPolicy.WARN_ONLY`. The suite also covers the omitted default, invalid enum, non-mapping block, and unknown nested key.

## DCO sign-off

## Author name correction

My full name is **Noah Ingwers** (`@noah-ing`). The original commit [`dac89aa`](https://github.com/agentrust-io/cmcp/commit/dac89aa5b6b589338585418b717810d1d0bf1d74) incorrectly used `Noah Ing` in its author, committer, and `Signed-off-by` fields. Those entries refer to me, Noah Ingwers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   ASCII-only records serialize to the same bytes as before, pinned by test, so nothing that verifies today stops verifying. #517 asks for JCS reuse precisely so a record can cross implementations, and nothing consumes these records at runtime yet, so the encoding is still free to correct.
 
+- **`catalog.drift_policy` was unreachable from configuration (#523).** The parser defined
+  the setting, but strict top-level validation rejected every `catalog:` block before it
+  could be read. The block and its only supported key are now explicitly allowlisted;
+  `fail_closed` remains the default, and misspelled nested keys fail closed.
+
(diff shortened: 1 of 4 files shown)


imran-siddique (approved):
Verified the dead-code path this reconnects. `load_config` already did `raw.get("catalog", {})` and built a `DriftPolicy` from it, but `_KNOWN_TOP_KEYS` omitted `catalog`, so strict validation rejected the block before the parser ever reached that line. The default worked only because it required no block at all, which means every operator who wrote an explicit `catalog:` stanza got a config error for a setting the proxy was already enforcing.

The unknown-nested-key guard is the part worth having. `test_unknown_catalog_key_raises` pins the misspelling case, so `drift_polciy: warn_only` is a config error rather than a silent fall back to `fail_closed` while the operator believes they set `warn_only`. That matches what `agent_manifest` and `kill_switch` already do, so the block reads like the rest of the file.

Default unchanged, no change to drift detection or comparison, nothing near the signing or attestation path. Merging.

Issue #518, opened 2026-08-16
Proposal: Hardened JSON‑RPC Gate for MCP Upstreams


tschecurity:
## Problem statement

The mock MCP upstream accepts arbitrary POST bodies, silently falls back on malformed JSON, returns 200 OK for invalid requests, and performs no validation of JSON‑RPC structure, tool names, or argument payloads. While each of these behaviors is individually harmless in a demo context, together they create an unbounded and inconsistent request surface. This makes it difficult to reason about upstream behavior, complicates debugging, and leaves the demo path without a clear security boundary. Anyone integrating or extending the mock upstream inherits these gaps.

This affects anyone using the mock upstream for testing, demos, or local development, especially when forwarding MCP requests from real runtimes or tools.

## Proposed solution

Introduce a **Hardened JSON‑RPC Gate**: a unified request‑validation layer that all MCP upstream requests must pass through. Instead of addressing each small weakness individually, the gate provides a single, coherent boundary that enforces:

**Payload safety**

- Maximum allowed Content‑Length
- Early rejection of oversized or partial bodies

**Strict JSON‑RPC validation**

- Reject malformed JSON (-32700)
- Require jsonrpc, id, and params (-32600)
- Validate arguments types (-32602)

**Capability and tool discipline**

- Validate tool_name as a string
- Optional allowlist for known tools
- Reject unknown or undeclared tools

**Argument safety**

- Require arguments to be a dict
- Cap key count, depth, and string length
- Reject non‑JSON‑safe types

**Unified error semantics**

- Never return 200 OK for malformed requests
- Always return proper JSON‑RPC error objects

**Security visibility**

- Log malformed or rejected requests
- Log capability violations
- Log oversized payloads

This creates a predictable, maintainable guardrail around the MCP request surface without changing runtime semantics or TRACE behavior.

## Alternatives considered

1. Patch each small issue individually  
Rejected because it produces a long tail of micro‑fixes without addressing the underlying need for a consistent security boundary. It also increases maintenance burden and makes behavior harder to reason about.

2. Replace the mock upstream entirely  
Rejected because the mock upstream is intentionally minimal and useful for demos. A hardened gate preserves its simplicity while adding safety.

3. Move validation into the runtime instead of the upstream  
Rejected because the upstream is the correct boundary for JSON‑RPC request validation. Runtime‑side validation would be redundant and harder to maintain.

## Security/TEE impact

None.
[...]

PR #544, opened 2026-08-21
fix(mock-upstream): hardened JSON-RPC gate for the mock MCP upstream

Yatsuiii:
Addresses #518

Adds a bounded JSON-RPC validation gate to the mock MCP upstream:

- rejects malformed JSON with -32700
- rejects structurally invalid requests with -32600
- rejects invalid tool names/arguments with -32602
- bounds request size before parsing
- logs rejected requests
- preserves existing valid-request behavior

The unmodified handler also crashed on non-object `params`; this is now covered by a regression test.

This intentionally does not close all of #518: strict jsonrpc/id validation, nested argument limits, non-standard JSON values such as NaN, and tool allowlisting remain follow-up scope.

## Verification

- 19 focused tests added
- unit suite: 1200 passed, 6 skipped, 6 pre-existing TPM-environment failures reproduced on unmodified `main`
- ruff clean
- mypy clean on the actual `src/cmcp_runtime/` package
- bandit: 0 findings

All required checks are clean relative to baseline; the six unit-test failures are pre-existing TPM environment failures reproduced identically on unmodified main.

### DIFF
--- a/scripts/mock_upstream.py
+++ b/scripts/mock_upstream.py
@@ -15,39 +15,121 @@ from __future__ import annotations
 
 import argparse
 import json
+import logging
 from http.server import BaseHTTPRequestHandler, HTTPServer
+from typing import Any
+
+# Matches MCPServer._max_request_bytes's default in cmcp_runtime/mcp/server.py,
+# so the mock rejects at the same boundary the real gateway does.
+MAX_REQUEST_BYTES = 1_000_000
+
+# How much of an oversized body we are willing to read and discard so the
+# client can finish writing it and read our 413 cleanly, instead of racing
+# an abrupt close (a client mid-write into a closed socket gets a
+# nondeterministic broken pipe, not the JSON-RPC error object this gate
+# promises). Bounded well above MAX_REQUEST_BYTES but still bounded, so an
+# attacker cannot turn this into the unbounded read the size check exists
+# to prevent -- past this ceiling we give up on a clean response and close.
+DRAIN_CEILING_BYTES = 10 * MAX_REQUEST_BYTES
+
+logger = logging.getLogger("mock_upstream")
 
 
 class MockMCPHandler(BaseHTTPRequestHandler):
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Reviewed against #518 and against the real server this mock stands in for.

The part that makes this worth having is that the boundary now matches. `MAX_REQUEST_BYTES = 1_000_000` is the same value as `MCPServer.max_request_bytes` on main at 296317c, so the mock rejects oversized bodies where the gateway does rather than at some other arbitrary point. A mock whose limits differ from the thing it substitutes for teaches you the wrong thing about your own system, which is the real cost of the gaps #518 lists.

The drain ceiling is reasoned rather than arbitrary, and the reasoning is right: read up to ten times the limit so a client mid-write can finish and read a clean 413, instead of racing an abrupt close and getting a nondeterministic broken pipe where this gate promises a JSON-RPC error object. Past that ceiling, give up on the clean response and set `close_connection`, so an unread tail cannot be misparsed as the start of the next request. That is the correct trade and both halves of it are bounded.

The tests spin up the actual `MockMCPHandler` on a random local port and speak raw HTTP to it, so they exercise the handler rather than a reimplementation of its rules. That matters more here than the count: a hand-rolled parser test would have passed against the old lenient handler too.

One behaviour change worth stating for whoever hits it: a request with no `params.name` used to answer `mock upstream: unknown called with {}` with a 200, and now gets -32602. That is the right direction, silent success on a malformed request is what made this hard to debug, but it is a change. Only `docker-compose.yml` and `docs/quickstart.md` drive the mock and both go through the gateway, which builds well-formed frames.

Non-blocking: this is demo scaffolding rather than shipped runtime code, so the value is boundary consistency and debuggability rather than closing an exploitable hole.

#518 stays open for the tool allowlist, argument depth and key-count caps, strict `jsonrpc` and `id` validation, and non-standard JSON values. Naming that scope in the PR rather than claiming the issue is the right way to have done this.

Merging.

PR #547, opened 2026-08-22
fix(catalog): judge approvals at the approval instant, not the wall clock (#533)

zohebk8s:
Closes the clock half of #533. Rebased on `main` now that #546 has landed, so this is the clock change alone.

## Why

`verify_catalog_change` judged `approved_at` and `expires_at` against `time.time()`. A record that was valid when the catalog was approved stopped verifying the moment its approvals expired, so the chain #517 exists to let an auditor replay could not be replayed. That makes the record an authorization token with a lifetime rather than a provenance record.

## What

**The interval is an assertion about when the signature could have been produced**, not a lifetime on the record. The caller passes `validity_instant`, a pinned checkpoint or transparency-receipt timestamp where it has one. Where it passes nothing, each approval is judged at its own `approved_at`, so the record stays verifiable indefinitely.

Requiring a pin instead would mean an auditor cannot verify a record without also holding the pin, which is a worse default than the one it replaces.

**`now` is gone with the wall clock**, and `time` is no longer imported. The rename is deliberate: the semantics changed, and a parameter still called `now` invites a caller to keep passing the verification clock into it.

**Revocation stays at verification time**, with the reason stated where the check is rather than only in the doc, since the two read as the same check to whoever touches this next:

```python
# Revocation is deliberately not judged at validity_instant. Whether a key was
# valid when it signed and whether it is revoked now are different questions.
```

The spec doc carries both rules, including that the interval bounds when the signature could have been produced, and now also states that the revocation list is load-bearing: with the interval no longer acting as a lifetime, revocation is the only thing that stops a historical approval from counting.

## Tests

```
tests/unit/test_catalog_approval.py     37 passed
with test_catalog_canonical_json.py     50 passed
coverage of approval.py                 148 statements, 0 missing, 100%
ruff / mypy --strict                    clean
full unit suite                         no new failures
```

The replay test is the one that would have failed before this change: approvals valid for `[100, 200)`, verified with no instant supplied, verified at 150, refused at `10**9`. A second pins that a revoked key is refused whatever instant the approvals are judged at.

Verified by mutation:

```
back to the wall clock              -> 2 failed, 35 passed
drop the window check               -> 3 failed, 34 passed
revocation at the approval instant  -> 2 failed, 35 passed
```

## One thing worth your eye

`TrustedReviewer` carries `principal_id`, `issuer`, `key` and `role`, and no validity window. So "reviewer key validity is checked against an instant inside that window" has nothing to check against beyond the approval's own interval, and that is what this implements: the instant governs the approval interval, not a key lifetime.


zohebk8s:
One consequence of the clock change is worth stating explicitly, since it is not a defect and I do not want it to arrive as a surprise later.

With the interval no longer acting as a lifetime, a record replays indefinitely. That means revocation is now the only mechanism that stops a historical approval from counting. Before this change, an approval signed by a key that later turned out to be compromised aged out on its own once `expires_at` passed, which was accidental rather than designed, but it was there. Now the revocation list carries that weight alone.

That follows from your decision rather than working against it, and it is the reason revocation stays at verification time. It does mean an operator's revocation list is load-bearing rather than a backstop, which is worth a sentence in the spec doc if you agree.

Related, and only as information: `TrustedReviewer` has no `not_before` or `not_after`, so nothing checks whether a key was valid at the instant it signed. Membership of `trusted_reviewers` plus revocation is the whole of the key-state check today. If that is the intended boundary, this PR is complete as it stands.

imran-siddique:
#546 is merged, so this needs a rebase onto `main` before it can be reviewed on its own terms.

Worth saying exactly why, because it is not a conflict you caused. I diffed the two branches against each other before merging: **all 139 added lines in #546 appear verbatim in this PR**, so #546 was a strict subset. Rebasing should drop the whole schema-wiring half and leave only the validity-clock change, roughly 71 lines.

What becomes redundant on the rebase, so nothing gets hand-reconciled twice:

- the entire `schemas/catalog-approval.schema.json`, `pyproject.toml` force-include and `scripts/verify_python_distribution.py` change, all landed in #546
- the `loader.py` and `test_catalog.py` edits, same
- the schema-authority section of `docs/spec/catalog-approval-provenance.md`, same, though your clock section is new and should stay
- most of the `test_catalog_approval.py` additions. The two that are genuinely yours and should survive are `test_a_record_stays_verifiable_after_its_approvals_expire` and `test_revocation_is_judged_now_not_at_the_approval_instant`

That leaves this PR as the answer to decision 1 in #533, which is the half that needs a maintainer ruling on which instant governs rather than a code review. Splitting it out this way makes that ruling reviewable on its own, which is the point.

Your note about revocation becoming the only mechanism that stops a historical approval is the right thing to have raised before it landed, not after. It goes in the spec doc alongside the clock change.

zohebk8s:
Rebased on `main`. Your read of the diff was right: git dropped the schema-wiring commit as already applied and the branch is now the clock change alone, 71 lines across 5 files, plus the doc commit below.

What the rebase removed matches your list exactly, nothing hand-reconciled:

```
schemas/catalog-approval.schema.json     not in the diff
pyproject.toml                           not in the diff
scripts/verify_python_distribution.py    not in the diff
src/cmcp_runtime/catalog/loader.py       not in the diff
tests/unit/test_catalog.py               not in the diff
```

The two tests you named are the two that survived in `test_catalog_approval.py`:

```
test_a_record_stays_verifiable_after_its_approvals_expire
test_revocation_is_judged_now_not_at_the_approval_instant
```

The doc keeps its clock section and does not re-add the schema-authority section.

The revocation point is now in the spec doc rather than only in a comment, as a second commit, beside the clock rule:

> Because a record replays indefinitely, revocation is the only thing that stops a historical approval from counting. An operator's revocation list is therefore load-bearing rather than a backstop: a key that should no longer count towards a threshold stops counting when it is revoked, and not before.

After the rebase: 50 passed across both catalog test files, `approval.py` at 148 statements with none missing, ruff and mypy strict clean.

imran-siddique (approved):
Reviewed against #533 and the rebased diff. The implementation matches the recorded clock decision: replay defaults to the signed approval instant, callers can pin an external validity instant, and revocation remains a verification-time check. The boundary and its revocation-list consequence are documented and covered by positive, boundary, replay, and revocation tests.

PR #550, opened 2026-08-23
ci: run conformance and integration tests

### LINKED ISSUE #524: ci: tests/conformance is never run, and it is currently red on main
## What

`.github/workflows/ci.yml:86` runs:

```
pytest tests/unit/ -v --tb=short --cov=src --cov-report=xml
```

`pyproject.toml:87` sets `testpaths = ["tests"]`, so a local `pytest` run covers everything. CI does not. `tests/conformance/` has never been executed by CI.

## Why it matters right now

`tests/conformance/test_gateway_conformance.py::test_mcp_initialize` currently fails on `main`:

```
assert result["protocolVersion"] == "2026-07-28"
E   AssertionError: assert '2025-11-25' == '2026-07-28'
```

The gateway is correct and the test is stale. Per #509, `initialize` belongs to the handshake era only, and `mcp/server.py:75` deliberately answers a legacy `initialize` with `_LEGACY_PROTOCOL_VERSIONS[0]`, which is `2025-11-25`. `PROTOCOL_VERSION` stays `2026-07-28` for the revision that removed `initialize` altogether. So the assertion encodes the pre-#509 behaviour.

The point is not the one stale assertion. It is that a conformance suite has been red on `main` for some time and nothing said so, because the job that would have said so does not run it.

## The shape of this

Same family as #521. A check exists, is reasonable to assume is running, and is not. A conformance suite is the worst place for it, since the entire value of one is that it runs and goes red.

## Suggested resolution

Two parts, and the second matters more than the first:

1. Fix the assertion to expect the negotiated legacy version for a legacy `initialize`, and add the counterpart asserting `PROTOCOL_VERSION` on the path that does not use `initialize`.
2. Put `tests/conformance/` in CI. If it is excluded deliberately, for runtime or flakiness, say so in the workflow with a comment and a tracking issue, so the next person does not read its presence in the tree as coverage.

Worth checking at the same time whether anything else under `tests/` is outside `tests/unit/` and therefore also unrun.

siiddhantt:
## What

- Correct the legacy `initialize` conformance expectation.
- Update the official-client fixture and assertion across the supported MCP 1.28.1 to 2.0 range.
- Run unit, conformance, and integration tests in CI. Keep `tests/soak/` explicitly outside per-PR CI.

## Why

CI currently runs only `tests/unit/`, leaving 59 conformance and integration tests outside required checks. Enabling those suites exposed a stale handshake assertion and an MCP 2.0 compatibility gap in the official-client fixture, so this fixes both in the same change that enables them.

Closes #524.

## Security impact

None. This changes test fixtures and CI coverage only.

## Test plan

- [x] `pytest` passes: 1,300 passed and 7 skipped on Windows with Python 3.13
- [x] Conformance and integration suites pass with MCP 1.28.1 and MCP 2.0
- [x] Focused suites pass on Python 3.11, 3.12, and 3.13
- [x] `ruff check` passes
- [x] `mypy` passes
- [x] Bandit and pip-audit pass
- [x] Manual before-and-after reproduction performed against upstream `main`

## DCO sign-off

cc @imran-siddique

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,8 @@ jobs:
         run: mypy src/cmcp_runtime/
 
       - name: Test
-        run: pytest tests/unit/ -v --tb=short --cov=src --cov-report=xml
+        # Soak tests are intentionally excluded from per-PR CI.
+        run: pytest tests/unit/ tests/conformance/ tests/integration/ -v --tb=short --cov=src --cov-report=xml
 
       - name: Upload coverage report
         uses: codecov/codecov-action@v7
--- a/tests/conformance/test_gateway_conformance.py
+++ b/tests/conformance/test_gateway_conformance.py
@@ -156,7 +156,8 @@ def test_mcp_initialize(client: TestClient) -> None:
     assert body["id"] == 1
     result = body["result"]
     assert "protocolVersion" in result
-    assert result["protocolVersion"] == "2026-07-28"
+    # `initialize` belongs to the legacy handshake era; the current revision removed it.
+    assert result["protocolVersion"] == "2025-11-25"
     assert "capabilities" in result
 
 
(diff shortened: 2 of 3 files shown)


imran-siddique (approved):
Reviewed against #524. The workflow now runs unit, conformance, and integration suites while explicitly excluding soak; the legacy initialize expectation matches the negotiated protocol; and the official-client compatibility changes are scoped to the test fixture and assertion. No production code changes.

PR #556, opened 2026-08-24
fix: close the remaining #518 gaps in the mock upstream's JSON-RPC gate

Yatsuiii:
Round two of #518, scoped by qubeena07's review comment on the issue.

## What this closes
- **Strict `jsonrpc`/`id` validation.** `jsonrpc` must equal `"2.0"` exactly. `id` must be a string, number, or null when present (bool explicitly excluded, since it's an `int` subclass in Python).
- **Argument depth and key-count caps.** `_MAX_ARG_DEPTH` and `_MAX_ARG_KEYS` follow the `_AIA_MAX_DEPTH` pattern in `tee/tpm.py`: named constants near `MAX_REQUEST_BYTES`, with a comment stating this is a judgment call rather than a value derived from an observed guarantee. The 1MB byte cap from #544 bounds total size, not shape — a payload well under that limit can still push toward Python's recursion limit through deep nesting, or cost real time to iterate through a flat object with thousands of short keys.
- **NaN/Infinity/-Infinity rejection.** `json.loads` takes these silently through its `parse_constant` hook by default; a custom `parse_constant` now raises, surfacing as the existing -32700 parse error path.

## What this does not touch
- **Tool allowlist**: still N/A, no catalog exists on the mock side (same reasoning as #544).
- **`_handle_mcp` in `mcp/server.py`**: has the same depth/key-count gap. qubeena07 flagged it but explicitly asked not to unify the two validation paths in this PR, so it's untouched here.
- **List length**: not capped. Scope here was "argument depth and key-count", matching what was asked for objects specifically; a large flat list is still bounded by the existing 1MB byte cap. Flagging in case that's worth its own pass.

## Tests
19 existing tests + 13 new (6 jsonrpc/id, 4 depth/key-count, 3 NaN/Infinity). All 32 pass. No changes outside `scripts/mock_upstream.py` and `tests/unit/test_mock_upstream_gate.py`.

### DIFF
--- a/scripts/mock_upstream.py
+++ b/scripts/mock_upstream.py
@@ -32,7 +32,48 @@ MAX_REQUEST_BYTES = 1_000_000
 # to prevent -- past this ceiling we give up on a clean response and close.
 DRAIN_CEILING_BYTES = 10 * MAX_REQUEST_BYTES
 
+# #518: the byte cap above bounds total size, not shape. A payload well
+# under 1MB can still be expensive to walk -- deep nesting pushes toward
+# Python's recursion limit, and a flat object with thousands of short keys
+# costs little in bytes but is not free to iterate. These are judgment
+# calls, not values derived from an observed cMCP tool call: no real
+# `arguments` payload we ship examples for goes past 3-4 levels or a
+# handful of keys, so both caps sit an order of magnitude above that,
+# leaving room for a legitimate tool with a genuinely nested schema.
+_MAX_ARG_DEPTH = 20
+_MAX_ARG_KEYS = 256
+
 logger = logging.getLogger("mock_upstream")
+
+
+def _reject_nan_and_infinity(text: str) -> float:
+    raise ValueError(f"non-standard JSON value not allowed: {text}")
+
+
+def _valid_rpc_id(value: Any) -> bool:
+    # JSON-RPC 2.0 id must be a string, number, or null -- not a bool, even
+    # though bool is an int subclass in Python.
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Merging this right after #561 so the pair lands together, which is what the comment-reference sync between the two constant sets depends on.

Most of what I would check here I already checked on #561: `_MAX_ARG_DEPTH = 20` and `_MAX_ARG_KEYS = 256` match exactly across both files, `json.loads("NaN")` returns `nan` on a default interpreter so the `parse_constant` gap is real, and @qubeena07's guidance on #518 asked for precisely this scoping, down to following the `_AIA_MAX_DEPTH` pattern in `tee/tpm.py` for how the constants are named and justified. You followed that pattern rather than approximating it.

The part I want to credit specifically is the comment on the caps saying this is a judgment call rather than a value derived from an observed guarantee. `_AIA_MAX_DEPTH` was sized from real observed data and these are not, and writing that down instead of borrowing the earlier constant's authority is the honest version. Someone tuning these later needs to know which kind of number they are looking at.

Flagging the uncapped list length rather than folding it in was also right. Scope was objects, a large flat list is still bounded by the 1MB cap, and widening quietly is how a reviewer's list stops being a list.

One note on what this is and is not. This is the demo mock, so its direct security value is lower than #561's. What it buys is that the mock now rejects what the real gateway rejects, which is the difference between a conformance reference and a lenient stand-in that lets a broken client look fine locally.

Approving and merging.

PR #561, opened 2026-08-24
fix: close #518's gateway-side gap in _handle_mcp

### LINKED ISSUE #518: Proposal: Hardened JSON‑RPC Gate for MCP Upstreams

Yatsuiii:
Closes the gateway-side gap qubeena07 flagged in review on #556 (round two of #518).

## What this closes

`_handle_mcp` in `src/cmcp_runtime/mcp/server.py` - the real inbound gateway path, not the demo mock - had the same gap #556 closed in `scripts/mock_upstream.py`:

- **Strict `jsonrpc`/`id` validation.** `jsonrpc` must equal `"2.0"` exactly. `id` must be a string, number, or null when present (bool explicitly excluded, since it's an `int` subclass in Python).
- **Argument depth and key-count caps.** `_MAX_ARG_DEPTH` and `_MAX_ARG_KEYS` are the same values as `scripts/mock_upstream.py`, with the same rationale in the comment: DOS-001's byte cap bounds total size, not shape, and a payload well under that limit can still push toward Python's recursion limit through deep nesting or cost real time to iterate through a flat object with thousands of keys.
- **NaN/Infinity/-Infinity rejection.** `json.loads` takes these silently through its `parse_constant` hook by default; a custom `parse_constant` now raises, surfacing as the existing -32700 parse error path (`MCP_PARSE_FAILURE`).

## What this does not touch

- **Unifying the two validation paths.** qubeena07 explicitly asked not to merge mock and gateway validation into shared code in this PR. The two now enforce the same values, kept in sync by comment reference rather than by shared code.
- **`arguments` must-be-a-dict type check.** The mock has this; the gateway doesn't, and it's a separate, unflagged gap - `_arg_shape_violation` tolerates non-dict `arguments` gracefully (walks it, no violation raised), so this PR doesn't regress anything by leaving it alone. Flagging in case it's worth its own pass.
- **Tool allowlist.** N/A here - unlike the mock, the gateway already gates tools downstream via the catalog and Cedar policy evaluation in `CMCPProxy.call_tool`.

## Refactor note

`_handle_mcp` and `_handle_tool_call` were both at or over this repo's cyclomatic complexity threshold before this change and would have gone further over it with the new checks inline. Extracted `_parse_mcp_envelope` (body size/parse, now returns the parsed dict or an error `Response`) and `_deny_response` (policy-deny error mapping) out of them - both are pure code motion, no behavior change, covered by the existing test suite passing unchanged.

## Tests

14 new tests (6 jsonrpc/id, 4 depth/key-count, 3 NaN/Infinity - matching #556's coverage shape) plus all 64 existing tests across `test_mcp_server_auth.py` and the other test files that instantiate `MCPServer` (`test_workflow_scope.py`, `test_break_glass.py`, `test_low_batch_186_187_191_194.py`, `test_session_reset.py`, `test_initialize_protocol_version.py`) pass. Ruff clean on both touched files.

### DIFF
--- a/src/cmcp_runtime/mcp/server.py
+++ b/src/cmcp_runtime/mcp/server.py
@@ -41,6 +41,47 @@ logger = logging.getLogger(__name__)
 # Endpoints exempt from bearer-token auth (Kubernetes liveness / readiness probes)
 _AUTH_EXEMPT_PATHS = {"/health", "/readyz"}
 
+# #518: DOS-001's byte cap bounds total size, not shape. A payload well under
+# the limit can still push toward Python's recursion limit through deep
+# nesting, or cost real time to iterate through a flat object with thousands
+# of short keys. These values mirror `_MAX_ARG_DEPTH` / `_MAX_ARG_KEYS` in
+# scripts/mock_upstream.py - same judgment call, same margin above any real
+# `arguments` payload this repo ships examples for, kept in sync rather than
+# unified per qubeena07's review on #518.
+_MAX_ARG_DEPTH = 20
+_MAX_ARG_KEYS = 256
+
+
+def _reject_nan_and_infinity(text: str) -> float:
+    raise ValueError(f"non-standard JSON value not allowed: {text}")
+
+
+def _valid_rpc_id(value: Any) -> bool:
+    # JSON-RPC 2.0 id must be a string, number, or null -- not a bool, even
+    # though bool is an int subclass in Python.
+    return value is None or (isinstance(value, (str, int, float)) and not isinstance(value, bool))
+
+
+def _arg_shape_violation(value: Any, *, depth: int = 0) -> str | None:
+    """Return a message describing the first depth/key-count violation found
+    under `value`, or None if it fits within `_MAX_ARG_DEPTH` / `_MAX_ARG_KEYS`."""
+    if depth > _MAX_ARG_DEPTH:
+        return f"arguments nested past the depth cap of {_MAX_ARG_DEPTH}"
+    if isinstance(value, dict):
+        if len(value) > _MAX_ARG_KEYS:
+            return f"object has more than {_MAX_ARG_KEYS} keys"
+        for child in value.values():
+            violation = _arg_shape_violation(child, depth=depth + 1)
+            if violation is not None:
+                return violation
+    elif isinstance(value, list):
+        for child in value:
+            violation = _arg_shape_violation(child, depth=depth + 1)
+            if violation is not None:
+                return violation
+    return None
+
+
 # Revisions the gateway can negotiate at `initialize`, newest first.
 #
 # `initialize` belongs to the handshake era only. `PROTOCOL_VERSION` (#509) is
@@ -271,8 +312,12 @@ class MCPServer:
             exception_handlers={Exception: _unhandled_error_handler},
         )
 
-    async def _handle_mcp(self, request: Request) -> Response:
-        """Handle MCP JSON-RPC 2.0 calls."""
+    async def _parse_mcp_envelope(self, request: Request) -> dict[str, Any] | Response:
+        """Read, size-check, and parse the request body.
+
+        Returns the parsed JSON-RPC message dict, or an error Response if the
+        body is oversized, unparsable, or not a JSON object.
+        """
         # DOS-001: reject oversized requests before parsing to prevent OOM
         content_length = request.headers.get("content-length")
         if content_length:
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
I read @qubeena07's guidance on #518 directly rather than taking your summary of it, and this matches it closely, including the parts it declines.

Their words: "I am not asking you to unify the two in this PR, but whatever caps you settle on here should eventually apply there too, otherwise the real gateway is left with the same depth and key count gap the mock just closed." That is this PR. Doing the follow-through in a separate change, on the file that actually matters, is the right shape.

Verified rather than taken:

- `json.loads("NaN")` returns `nan` on a default interpreter, so the `parse_constant` gap is real. Routing the rejection through the existing `ValueError` branch so it surfaces as `-32700` / `MCP_PARSE_FAILURE` is better than adding a new error path for it, because a caller sees one parse-failure contract rather than two.
- Constant parity is exact. `_MAX_ARG_DEPTH = 20` and `_MAX_ARG_KEYS = 256` on both sides.
- The `_deny_response` extraction you describe as pure code motion is. The `upstream_error:` 502 branch, the `{attestation_stale, catalog_drift}` 503 branch and the `-32000` codes are identical between `origin/main` and this head, relocated and not rewritten. That was the part I was going to look hardest at, since a refactor landing in the same commit as new checks is where behaviour usually goes missing quietly.

Two judgement calls I want to endorse explicitly, because both could have gone the lazy way.

Excluding `bool` from the valid `id` types. It is an `int` subclass in Python, so the obvious `isinstance(id, (str, int, type(None)))` accepts `true` as a request id and nothing downstream would notice. Naming it in the description rather than leaving it as a silent detail is what let me check it quickly.

Leaving the `arguments` must-be-a-dict gap alone and saying so. `_arg_shape_violation` tolerates a non-dict gracefully, so you are right that not touching it regresses nothing, and flagging it beats folding an unflagged change into a PR scoped to a reviewer's list.

One note for the record rather than for you: the caps are kept in sync by comment reference rather than by shared code, so that reference depends on #556 landing. Taking that next.

Approving and merging.

Issue #432, opened 2026-07-30
security(tee): the gateway is never measured into a PCR, so the policy tamper claim is unenforced


imran-siddique:
The TPM provider reads PCRs 0 through 7 and reports `sha256(PCR0 || ... || PCR7)`. Those PCRs cover firmware, option ROMs, boot configuration, and bootloader. They do not cover the cMCP gateway binary, its policy bundle, or its effective configuration. There is no `PCR_Extend` call anywhere in the codebase.

**Impact.** The stated purpose of the TPM path, protecting policy from tampering inside the gateway, is not enforced by the TPM. Replacing the policy bundle or the gateway binary produces an identical measurement.

**Fix, with an important caveat.** The obvious approach is to extend an application PCR at startup with a digest over binary, policy bundle, and config, and include it in the quote selection. Per the TCG PC Client Platform TPM Profile, PCR 23 is Application Support and PCR 16 is Debug, and both are resettable from locality 0. An adversary with local code execution can reset PCR 23 and re-extend a chosen value, so this alone is advisory rather than tamper-resistant.

Options to evaluate in the RFC:
- NV extend index (`TPM_NT_EXTEND`) with a write policy, which is not resettable from locality 0.
- Sealing the TRACE signing key to a policy over both the application PCR and non-resettable SRTM PCRs, so a reset breaks unseal.
- DRTM where the platform supports it.

Part of the TPM security model RFC.

imran-siddique:
Tracked in the TPM security model RFC: #439.

imran-siddique:
**Decision: NV extend index, not an application PCR.**

Rationale, from the TCG PC Client Platform TPM Profile: PCR 23 is Application Support and PCR 16 is Debug, and both are resettable from locality 0. An adversary with local code execution can reset PCR 23 and re-extend a value of their choosing, so measuring the gateway there is advisory rather than tamper-resistant against precisely the adversary this tier exists to address.

An NV extend index (`TPM_NT_EXTEND`) with a write policy is not resettable from locality 0, so it survives that attack.

Implementation notes for whoever picks this up:

- Define the index with `TPM_NT_EXTEND` and a write policy, extended at gateway startup with a digest over the binary, the policy bundle, and the effective configuration, before serving traffic.
- Include it in the evidence alongside the quote. Note it is an NV read rather than part of the PCR selection, so it needs its own field and its own freshness argument, unlike a PCR that the quote covers directly.
[...]

PR #563, opened 2026-08-24
fix(tee): bind the gateway measurement into SEV-SNP and TDX report_data

### LINKED ISSUE #552: tee: bind the gateway measurement into SEV-SNP and TDX report_data, refreshed on policy reload

zohebk8s:
Closes #552. Both bullets of the proposed change, wired for the three providers the issue names. The report the binding lands on is not the one a session claim carries; see *Not covered*.

## Why

`gateway_measurement()` folds installed code, the policy bundle and the effective config into one digest. #432 extends it into a `TPM_NT_EXTEND` NV index and certifies it with a `TPM2_NV_Certify` pair, validated on real Azure Trusted Launch vTPM hardware. The "what to measure" question is answered and proven.

It ran for the `tpm` provider only. At `startup.py:160`, `_measure_gateway` returned early on `tee_provider.provider_name() != "tpm"`.

The docstring's reason was that SEV-SNP and TDX commit their own binding through the report's fields. That is true and it is not equivalent. **Those fields carry the launch measurement, which is boot-time.** It does not move when the Cedar bundle reloads mid-session via `PolicyEvaluator._maybe_reload()`. So on SEV-SNP, TDX and Azure CVM, the policy actually in force was committed to nothing.

## What

No new commitment scheme. The existing, already-validated digest goes into those platforms' `report_data`.

**`make_measurement_bound_nonce(tee_public_key, measurement_digest)`** in `tee/base.py`, using the layout `make_audit_bound_nonce` already uses:

```
jwk_thumbprint(pubkey) (32) || measurement_digest (32)
```

`gateway_measurement().digest` is already a raw 32-byte SHA-256, so it drops into the second half unreshaped and a verifier compares it against a digest it recomputes rather than against a hash of one. It sits beside `make_audit_bound_nonce` and reuses the same `jwk_thumbprint` helper, so the first half is literally the same code path.

**Applied to `sev-snp`, `tdx` and `azure-cvm-sev-snp`.** All three place the whole 64-byte nonce in `report_data`. SEV-SNP and TDX write it straight into REPORT_DATA and REPORTDATA. Azure CVM cannot, since the paravisor owns SNP REPORT_DATA, and commits `sha256(nonce)` into the AK-signed quote instead, which is what `cmcp_verify.azure_cvm` already re-derives. `tpm` keeps the random salt: its NV index has an append-only history that `report_data` does not.

**Freshness survives losing the salt.** The gateway generates a new signing key on every start, so `report_data[:32]` still differs between two starts of byte-identical code, policy and config.

**Refreshed on every policy-bundle reload**, not only the reloads that moved the hash. `report_data` holds one value and no history, so nothing in a report says whether it is current: the same digest signed now and that digest signed an hour ago are different assertions, and only the latest report reaches a verifier. `PolicyEvaluator._maybe_reload` calls a new `on_reload` hook whenever `PolicyStore.reload_if_stale` reports the bundle was re-read. The cost is bounded by `policy_reload_interval_seconds` rather than by request rate, because `reload_if_stale` stamps its clock before the attempt, and it is zero in the default configuration where reloading is off.

**Recompute-and-compare is a check, not a note in the spec.** `verify_trace_claim` gains an optional `expected_gateway_measurement` (raw 32 bytes, hex, or `sha256:`-prefixed) and a `MEASUREMENT_NOT_BOUND` failure reason, as step 7c. Opt-in, because the expected digest has to be an out-of-band trust input like `ApprovedHashes`: a check that read it out of the claim would be asking the claim to vouch for itself.

## Tests

40, software-only by design. `SoftwareOnlyProvider` echoes the nonce into `report_data`, which is the collector-side shape SEV-SNP and TDX give, so the round trip proves the contract and leaves the silicon question where the issue leaves it.

The five properties prototyped on the issue: a deterministic measurement across repeat calls, a one-character Cedar change moving the policy component and only the policy component, a full round trip whose independent recompute matches both halves of `report_data`, a mid-session reload producing a report that reflects the new measurement, and a stale pre-reload measurement correctly rejected on recompute.

Verified by mutation, twelve guards altered one at a time, each killed. Restoring the narrower "refresh only when the hash changed" reading fails 2, re-hashing the digest verifier-side fails 3, and reverting to the tpm-only early return fails 2.

Three checks exist because the change was audited against the issue text rather than against itself:

- the reload hook was only ever exercised with `reload_if_stale` stubbed, so a real `PolicyStore` over a real on-disk bundle now reloads an **unchanged** bundle and asserts the hook still fires, which is the second bullet with nothing mocked on the seam
- `startup._jwk_thumbprint_sha256` and `tee.base.jwk_thumbprint` each build `report_data[:32]` and nothing forced them to agree, so a test does
- all three providers are asserted to place the whole 64-byte nonce in `report_data`, since the binding is only real if it reaches the field the hardware signs

Full suite: 1242 passed, 9 skipped. Ruff and mypy clean. Nine failures are pre-existing and unrelated, all from the installed `agent_manifest` SDK in my environment (`module 'agent_manifest' has no attribute 'COSE_MANIFEST_VERSION'`); `tests/unit/test_agent_manifest_cose.py` and `tests/unit/test_intent_binding.py` fail to collect for the same reason.

## Decisions worth reviewing

**An unmeasurable gateway is now fatal at startup on these three providers**, in production, as it has been on `tpm` since #432. Extending the measurement to a platform extends the consequence of not having one. `CMCP_DEV_MODE=1` still downgrades it to a warning, which is what an editable install with no `RECORD` metadata needs. This will stop an existing SEV-SNP deployment that runs an editable install without dev mode, so it is the one behaviour change here.

**A refresh that cannot re-attest is fail-open**, unlike startup. It logs and keeps the previous report rather than refusing traffic. Failing closed there would trade a detectable weakness for an outage: stale `report_data` no longer matches the recomputed digest, so a verifier rejects the claim. This is the same trade AUDIT-006 already makes for a failed per-session attestation in `SessionManager.create_session`.

**Re-attesting on an unchanged bundle is deliberate**, and it is the literal reading of the second bullet. The narrower reading, refreshing only when the hash moved, would leave the last report standing for as long as policy happens not to change, which is exactly the staleness the absent history cannot express.

**`ctx.gateway_measurement` and `ctx.attestation_report` are two assignments, not one atomic swap.** Neither is touched until the new report is in hand, so a failure at any earlier step leaves both unchanged, but a concurrent reader could see the new measurement beside the old report for one interpreter step. Benign today because nothing reads the pair together; the docstring records it rather than overclaiming.

## Not covered

- **The binding is on the gateway's startup report.** A TRACE Claim for a session carries the per-session report where one was produced, and AUDIT-006 already commits the audit-chain root in the same `report_data[32:64]`. So on the normal session path a verifier sees the chain-root commitment, not the measurement, and step 7c applies to claims that fall back to the startup report. Both cannot occupy one 64-byte field as the layout stands. #552 scopes the audit-chain-root binding out, so this is left for the issue that takes that on rather than resolved either way here.
- **Hardware validation on real SEV-SNP and TDX silicon**, out of scope per the issue. The contract lands with software-only proof.
- **Unrelated, noticed in passing:** `_VALID_PROVIDERS` in `startup.py` is missing `azure-cvm-sev-snp`, though the value is present in both `_ALLOWED_PROVIDERS` and `_PROVIDER_MAP`. A gateway that auto-detects the Azure confidential-VM provider hits `ATTESTATION_PROVIDER_INVALID` and exits. Not touched here; happy to raise it separately.


imran-siddique (approved):
Reviewed this properly rather than on the description, because it changes what the hardware signs.

**The gap is real and stated correctly.** SEV-SNP and TDX committing their own binding through the report's fields is true and is not equivalent, and the reason is exactly the one you give: those fields carry a launch measurement. It is fixed at boot and does not move when `PolicyEvaluator._maybe_reload()` swaps the Cedar bundle. So the policy actually in force on those three platforms was committed to nothing, while the `tpm` tier has had this since #432.

**The layout choice checks out.** `make_audit_bound_nonce` at `tee/base.py:117` is `jwk_thumbprint(pubkey) || SHA-256(chain_root)`; `make_measurement_bound_nonce` at :135 is `jwk_thumbprint(pubkey) || measurement_digest` through the same helper. Putting the raw digest in rather than a hash of it is the detail that matters, because it lets a verifier compare against a digest it recomputes instead of against a commitment it cannot open. And replacing the random salt without losing freshness holds for the reason you give: the signing key is per start, so `report_data[:32]` still differs across two starts of byte-identical code, policy and config.

**The collision was the thing I came here to raise, and you had already found it.** Both commitments target `report_data[32:64]` with no domain separation between them. Your *Not covered* says it plainly, scopes it to the issue that takes on the audit-chain-root binding, and does not pretend the layout resolves it. That is the right call. Recording it as a live constraint is worth more than a partial fix would have been.

**Verified rather than taken:** ran `tests/unit/test_measurement_report_binding.py` on a detached worktree, 50 passed, then applied your named mutation reverting `_measure_gateway` to the `tpm`-only early return. Exactly 2 failures, `test_startup_binds_the_measurement_into_the_nonce_on_sev_snp` and `test_an_unmeasurable_gateway_aborts_startup_on_sev_snp`, which is what your table claims. The mutation numbers are honest.

Three things I want to name because they are decisions, not details.

Re-attesting on an unchanged bundle is right, and your reasoning for it is the load-bearing part: `report_data` holds one value and no history, so "the digest signed now" and "the digest signed an hour ago" are different assertions that the field cannot distinguish. The narrower reading would leave a report standing for as long as policy happens not to change.

Fail-open on refresh against fail-closed at startup is the right asymmetry, and it is safe for the reason you state rather than by luck: stale `report_data` stops matching the recomputed digest, so a verifier rejects it. The weakness is detectable, and refusing traffic instead would trade that for an outage.

Making an unmeasurable gateway fatal on three more platforms is the one behaviour change here, and calling it out under *Decisions worth reviewing* rather than burying it in the changelog is what let me agree with it quickly.

**On `_VALID_PROVIDERS`, thank you, and I am taking it rather than leaving it with you.** Confirmed: `startup.py:57` holds five names without `azure-cvm-sev-snp`, `audit/trace_claim.py:25` holds six with it, and the comment on line 55 says the first mirrors the second. So `startup.py:420` raises `ATTESTATION_PROVIDER_INVALID` and this PR adds binding for a provider that path cannot reach. Follow-up from me, with a test asserting the two sets agree so they cannot drift apart again.

Approving and merging.

Issue #565, opened 2026-08-24
Define session-independent execution correlation for post-session MCP


imran-siddique:
Track the MCP 2026 compatibility gap identified in the roadmap impact review. Define a cMCP execution identifier that survives independent requests without implying an MCP protocol session. Acceptance evidence: protocol-version vectors; correlation across retries and multi-request work; explicit collision, replay, and missing-context failures; TRACE records remain offline-joinable.

Yatsuiii:
I wrote the design this issue now builds against in #574, so here is the implementation shape I would take it in, and a scope boundary I would want checked before starting.

**Correlation and durable state.** An executable state machine over admission, reservation, invocation, response, audit persistence, retry, recovery and terminal disposition. One authoritative durable record per `execution_id`, keyed under the authenticated agent identity as ruled on #574. The identifier and the immutable action digest reserved atomically before invocation. Terminal state and audit evidence commit together, or recovery leaves the execution blocked and unable to authorize a replay. Tests for concurrent collision, process restart, persistence failure, mutated action, missing context and terminal replay.

That slice proves only that independent requests can be correlated. It does not make retry safe, and I would not present it as exactly once execution.

**Then the three conditions that share the boundary.** #567's Task transitions and authenticated callbacks, #568's split between workload identity, represented user and delegated authority, and #566's catalog checkpoint binding. Each has to preserve the same identity boundary rather than grow its own version of it, which is the part that argues for doing them as one sequence. On #566, @solloek369-arch has proposed keeping the discovered schema, the approved checkpoint and the checkpoint attributed to a governed call independently mutable. That is the right shape for the acceptance matrix and I would build against it rather than around it.

**Falsification before claims.** A stateful harness that generates valid sequences then mutates ordering, identity, action digest, policy hash, tool digest, callback authority and persistence outcome, with crashes injected before and after each durable boundary. Every failure has to produce a stable finding or terminal disposition, never an uncaught exception or a silent evidence gap.

**Scope question.** I would leave HTTP-native transport out. It is P1 in the roadmap impact document with no issue filed, and #574's own reasoning was that the `params._cmcp` envelope "keeps the design neutral across transports and avoids depending on future HTTP header design." Same test applied here. If you intend to file it during this work, that changes the scope and I would rather know before starting than after.

I would like to take this sequence: #565 first, then #567, #568 and #566 as each boundary lands, in that order rather than in parallel. If any of it is already spoken for, say so and I will stay out of the way.

[...]

Issue #571, opened 2026-08-25
Post-execution exceptions leave a completed tool call with no audit outcome


Yatsuiii:
## Summary

`CMCPProxy.call_tool` writes a terminal audit entry for an unexpected exception raised **before** the upstream tool call, and writes nothing for one raised **after** it. Since the tool has already run by then, an upstream can execute a tool call and leave no record of it in the audit chain.

The invariant is not missing from this codebase. It is implemented on one side of the upstream request only.

`POLICY-003` (`tests/unit/test_mcp_proxy.py:556`) asserts that a Cedar backend exception appends a `fault` entry before re-raising. That is the pre-execution case, and it works. There is no equivalent boundary on the response side, and everything between `_forward_to_upstream` returning and the audit append at `proxy.py:1221` is fallible.

Verified end to end against `main` at `a2893da`. The reproduction below exits 0 on confirmation and includes `POLICY-003` as a live control in the same run.

## Reproduction result

```
--- A  nested text inside a valid envelope
    crash path:
        cmcp_runtime/mcp/proxy.py:1209  call_tool()
        cmcp_runtime/mcp/proxy.py:139  _extract_external_execution_evidence()
    upstream invoked    : ['test.echo']
    outcome             : RecursionError propagated out of call_tool
    audit entries       : 1 -> 1  ['session_start']
    chain tip changed   : False

--- B  nested envelope, before the size guard runs
    crash path:
        cmcp_runtime/mcp/proxy.py:975  call_tool()
        cmcp_runtime/mcp/proxy.py:525  _forward_to_upstream()
        cmcp_runtime/mcp/streamable_http.py:113  parse_response()
    upstream invoked    : ['test.echo']
    outcome             : RecursionError propagated out of call_tool
    audit entries       : 1 -> 1  ['session_start']
    chain tip changed   : False

--- C  control: pre-execution exception (POLICY-003)
    crash path:
        cmcp_runtime/mcp/proxy.py:861  call_tool()
    upstream invoked    : []
    outcome             : RuntimeError propagated out of call_tool
    audit entries       : 1 -> 2  ['session_start', 'fault']
    chain tip changed   : True
```

Same class of unexpected exception. One side of the upstream request records it, the other does not.

## The concrete trigger

`json.loads` raises `RecursionError` on deeply nested input. `RecursionError` subclasses `RuntimeError`, not `ValueError`, so it is caught by neither of the two guards standing between the upstream call and the audit write:

[...]

Issue #568, opened 2026-08-24
Separate workload identity, represented user, and delegated authority claims


imran-siddique:
Define separate cMCP claims for workload identity, represented user, token audience, proof-of-possession binding, and delegation reference. Map DPoP and token exchange without treating authentication as runtime attestation. Acceptance evidence: confused-deputy, token replay, audience mismatch, and identity/attestation mismatch tests.

imran-siddique:
The next deliverable is the identity/authority claim boundary and the confused-deputy, replay, audience, and attestation-mismatch acceptance matrix. Keep this independent from #565’s action identity and #566’s catalog identity.

athena-kanellatou:
I traced the current identity path, and I think the cleanest first step is to define the identity/authority boundary and acceptance matrix before choosing a wire format.

The runtime already has a workload identity path through agent_id, authenticated_subject, and subject_source. I would keep user/delegation authority separate from that rather than overloading authenticated_subject.

A possible boundary is:

1. workload identity — which agent/workload is making the request

2. represented user — the principal the workload is acting for, when present

3. token audience — the resource the authority is intended for

4. proof-of-possession binding — the key binding associated with that authority, where applicable

5. delegation reference — the evidence that carries authority from the represented principal to the workload

6. runtime attestation — evidence about the executing workload/runtime, never inferred from authentication alone

For the acceptance matrix, I would expect at least these cases:

Valid workload identity + valid authority + consistent audience/bindings
Expected: accept subject to policy

Authority has the wrong audience
Expected: refuse before upstream invocation

Authority is presented outside its required PoP/workload/delegation binding
Expected: refuse

Delegated authority is replayed outside its valid scope or context
Expected: refuse

Represented-user or delegation evidence conflicts with the authenticated workload identity
Expected: refuse rather than merge the identities

Authentication succeeds but runtime attestation is absent or unverified
Expected: authentication may still be valid, but no runtime-attestation claim

No represented user is present
Expected: do not synthesize one, instead evaluate authorization from workload authority and policy

For conformance, I think each negative case should identify which boundary failed, rather than collapsing workload authentication, delegated authorization, PoP, and attestation into one generic identity error.

I have intentionally not proposed field names or a token-exchange wire format yet, since those seem downstream of agreeing on the boundary.

If this matches the intended direction, I’m happy to turn it into the concrete acceptance matrix/conformance cases first and implement the agreed claim representation separately.

imran-siddique:
[...]

Issue #567, opened 2026-08-24
Bind MCP Tasks and authenticated callbacks into cMCP evidence


imran-siddique:
Define how task creation, updates, cancellation, callbacks, and terminal results bind to one cMCP call graph and policy context. Acceptance evidence: replay and out-of-order callback tests; cancellation-authority checks; terminal result binding; documented limits for delivery and completion claims.

solloek369-arch:
One additional boundary from final SEP-2663 may be useful alongside the identity/authority boundary already called out for #567 on #565.

The Tasks lifecycle gives deliberately different evidentiary strength to different successful responses:

Task creation: a server MUST NOT return CreateTaskResult until the task is durably created and a subsequent tasks/get for that taskId would resolve.
Task update: success is an empty acknowledgement and is explicitly eventually consistent. The acknowledgement may precede any observable state change, and input for unknown, already-satisfied, or superseded request keys may be ignored.
Task cancellation: success is also an empty acknowledgement, but cancellation is cooperative. The request signals cancellation intent; the acknowledgement does not establish that cancellation took effect, and the task may legitimately reach a terminal state other than cancelled.

That suggests a separate evidence rule for the task lifecycle:

The evidentiary strength assigned to a protocol response should not exceed the lifecycle guarantee that the protocol actually gives that response.

Or, for the two control methods:

request acknowledgement ≠ applied state transition ≠ terminal task outcome

The creation path is a useful positive control because SEP-2663 intentionally gives its response a stronger guarantee: durable task existence has already been established before the response is returned.

A small differential set could therefore be:

CreateTaskResult returned → the task already resolves through tasks/get;
tasks/update acknowledged → the acknowledgement alone does not establish input consumption or an observable state transition;
an already-satisfied or superseded input key may produce the same successful acknowledgement while being ignored;
valid case: tasks/cancel acknowledged → task later reaches cancelled;
contrasting valid case: the same cancellation acknowledgement → task legitimately reaches another terminal state because cancellation did not take effect before completion.

For evidence, that seems to argue for keeping at least:

control request issued
control request acknowledged
state transition observed
terminal state established
external effect evidenced, if separately available

independently attributable, so that evidence at one lifecycle layer cannot silently stand in for a stronger claim at the next.

imran-siddique:
[...]

PR #574, opened 2026-08-25
docs: propose session-independent execution correlation (#565)

Yatsuiii:
## What

Records the agreed design for #565 in `docs/spec/execution-correlation.md`.

The document keeps the current identity finding prominent. `initialize` creates no client negotiated identity. `session_id` and `call_id` are gateway generated. `workflow_id` is supplied by the caller but is not a validated execution identity.

## Settled decisions

1. `execution_id` is a typed `AuditEntry` field, not a `detail` key.
2. Collision within one authenticated agent identity is refused.
3. The TRACE Claim does not enumerate `execution_id` values.
4. Reuse after a terminal outcome is refused with no replay window.
5. `execution_id` and `workflow_id` remain independent.

## Claim boundary

The document records the boundary raised in trace-spec#66:

`attempt identity ≠ logical operation identity ≠ external outcome identity`

A shared `execution_id` establishes correlation only. Treating two requests as the same logical operation also requires an immutable canonical action or intent binding. A different binding under the same identifier is a collision or mutated operation and must be refused. This change does not invent another schema field for that binding.

## Correctness requirements

Collision and replay enforcement requires authoritative execution state keyed by authenticated agent identity and `execution_id`. Reservation must occur atomically before upstream invocation, and terminal state must be durable before later replay classification.

## Scope

Documentation only. No running behavior or schema changes are made in this PR.

## Verification

1. `git diff --check` passes.
2. Ruff passes for `src/cmcp_runtime`, `src/cmcp_verify`, and `tests`.
3. MyPy passes for 64 source files.
4. All cited source locations were checked against the current branch.
5. The complete local test command stalled in gateway conformance and was stopped. GitHub CI remains required for the final verdict.

## DCO sign off

I certify that I wrote or have the right to submit this contribution and agree to the Developer Certificate of Origin.

### DIFF
--- a/docs/spec/execution-correlation.md
+++ b/docs/spec/execution-correlation.md
@@ -1,0 +1,135 @@
+# Session-Independent Execution Correlation
+
+Status: proposal for [#565](https://github.com/agentrust-io/cmcp/issues/565), not
+adopted. This document does not change the cMCP protocol.
+
+[mcp-2026-roadmap-impact.md](mcp-2026-roadmap-impact.md) lists this as P0 and
+first in the issue sequence: "Replace dependence on MCP initialization/session
+lifetime with a cMCP execution identifier carried across requests. A TRACE claim
+must remain joinable without implying an MCP protocol session."
+
+## What binds evidence today
+
+Worth stating precisely, because the dependence is not where the phrase
+"MCP session" suggests.
+
+`initialize` does not create anything. `MCPServer._handle_mcp` answers it at
+`server.py:449` by negotiating a protocol version and returning capabilities. No
+session is minted, no `Mcp-Session-Id` is issued, and nothing about the handshake
+is carried into evidence.
+
+The binding is entirely gateway-side:
+
+| Identifier | Origin | Scope |
+|---|---|---|
+| `session_id` | `SessionManager.create_session`, `session/manager.py:118`, a `uuid4` | One gateway session |
+| `call_id` | `server.py:569`, a fresh `uuid4` per `tools/call` | One tool call |
+| `workflow_id` | caller-supplied via `params._cmcp.workflow_id`, `server.py:571-575` | Caller-defined |
+
+`session_id` is what everything joins on. Audit entries carry it, the TRACE Claim
+is issued per session on close and stored keyed by it (`manager.py:182`), and the
+read paths are `/sessions/{session_id}/trace-claim` and
+`/audit/export?session_id=` (`server.py:341`, `server.py:705`). At session
+creation the audit chain root is bound into TEE `report_data` where the platform
+supports it, and `manager.py` warns explicitly when it could not be.
+
+So a cMCP session is a gateway lifetime with a hardware-anchored chain root and
+one claim at the end. It is not an MCP protocol session and never was.
+
+## The gap
+
+The problem is not that sessions disappear from the protocol. It is that
+`session_id` is the only cross-request correlation key, and it carries two jobs
+that stop being compatible once work arrives as independent requests:
+
+1. It scopes the evidence bundle, one chain root, one claim.
+2. It is the only thing relating two calls that belong to the same intent.
+
+With independent requests there are two options and both fail:
+
+**One long-lived session.** Correlation works, but the TRACE Claim grows without
+bound and evidence is not available until close. A claim that never closes is not
+evidence anyone can act on.
+
+**A session per request.** Claims stay small and prompt, but every call is its
+own chain root and nothing relates two calls. Multi-request work becomes
+unjoinable, which is exactly what the roadmap requires stay joinable.
+
+`workflow_id` is closest to prior art here, but it is caller-supplied and
+unverified, so it cannot carry correlation on its own.
+
+There is a concrete failure already filed. In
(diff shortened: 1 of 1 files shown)


imran-siddique:
Ruling on all five. You were right not to settle them alone, and right that a wrong shape here propagates into #566, #567 and #568.

First, the finding, because it is the part that stands regardless of what happens to the proposal. Verified at pinned `21df5b42`: `initialize` at `server.py:449` negotiates a protocol revision and mints nothing, and `manager.py:118` is `session_id = str(uuid4())`, gateway-side. So "MCP session" suggests a client-negotiated identity that does not exist. That is the same discovery your #571 thread reached from the other direction, where `call_id` is also a fresh server-side `uuid4`. Three identifiers, none of them correlatable by the caller. Write that down in the document even if everything below changes.

## 1. Field on `AuditEntry`, not `detail`

A field. The schema change is the smaller cost.

`detail` is where things go when we have not decided what they are. Something the verifier must key on, and that #567 and #568 will both consume, is not that. Putting a correlation identifier in a free-form dict means every consumer re-derives its own expectations about presence, type and absence, and they will not agree.

The append-only hashed structure argues for a field rather than against one: a typed field that is always present, even as null, hashes predictably, while a key that appears in `detail` only sometimes changes the shape of what is hashed depending on a value's presence.

## 2. Collision: refuse

Refuse, inside a single agent identity.

A correlation identifier whose collisions are recorded rather than refused is not an identifier, it is a hint, and everything built on it inherits the ambiguity. #568 in particular cannot do anything useful with "these two executions might be the same one".

Refusing is also the honest failure: a caller that reuses an `execution_id` has a bug or is an attacker, and both should be told immediately rather than discovered by an auditor later.

## 3. No, the TRACE Claim does not enumerate them

An enumeration is unbounded, grows with session length, and duplicates what the audit chain already orders. The Claim binds the chain; the chain carries the identifiers.

There is also a disclosure argument: a Claim is the artifact most likely to be shared outside the operator, and a list of execution identifiers leaks work volume and shape to a reader entitled only to verify one record.

## 4. "Refuse reuse after a terminal outcome" is enough. No replay window.

A window is a clock, and a clock is a second source of truth that has to be synchronised, configured and reasoned about at verification time years later. We have just spent a week on what happens when a verifier's clock disagrees with an issuer's.

Terminal outcome is already the boundary this system understands. #571 landed `not_attempted` and `outcome_unknown` precisely to make that boundary explicit, so key the refusal on it and inherit the work rather than introducing a parallel notion of expiry.

One consequence to write down: an execution that never reaches a terminal outcome is never reusable and never expires. That is the fail-closed direction and it is correct, but it means a stuck execution holds its identifier permanently, and someone should know that before they see it.

## 5. Independent of `workflow_id`

They answer different questions. `workflow_id` groups work that belongs together by intent; `execution_id` identifies one unit of work for correlation across requests. Deriving either from the other collapses two questions into one and the collapse is not reversible.

Independent, with the relationship documented rather than encoded. If it turns out every `execution_id` in practice has exactly one `workflow_id`, that is an observation worth recording, not a constraint worth enforcing.

## What I would like

Fold these into the document as decided rather than open, keep the `initialize`/`uuid4` finding prominent, and mark it agreed rather than proposed. Then it is the thing #566, #567 and #568 build against.

Merging once that lands. Thanks for opening it for argument instead of implementing a guess.

imran-siddique (approved):
All five landed, and you carried the consequences rather than just the verdicts. That is the difference between a document that records a decision and one that can be built against.

Two places where that shows.

**On the replay window.** I said key the refusal on terminal outcome and inherit #571's work. You wrote:

> This rule includes `outcome_unknown`: uncertainty must not become permission to repeat an irreversible effect.

That sentence is doing more than the rule I gave you. `outcome_unknown` is exactly the state where a retry layer is most tempted to try again, and saying plainly that uncertainty is not permission closes the reading where someone treats "we do not know" as "so it probably did not happen".

**On the cost of that choice.** I asked you to write down that a stuck execution holds its identifier forever, and you did, including the part I had not said:

> A stuck execution therefore holds its identifier until an explicit recovery design exists. This is the fail closed result.

Naming that a recovery design does not exist yet, rather than leaving the reader to discover it, is what stops this being rediscovered as a bug in six months.

**Two additions I did not ask for and would have wanted.**

Scoping collision so the same value under different agent identities does not collide. My ruling said refuse within one agent identity and I did not say what happens across identities; leaving that implicit would have produced a verifier that refuses legitimate values from unrelated agents.

And a missing `execution_id` recording `null` in the typed field rather than synthesizing one, which gives consumers the property that matters: "a present value came from the validated caller path." A synthesized identifier would have looked identical to a real one, which is the same class of defect as the reported-but-unenforced fields we have been fixing all week.

Keeping the reasoning on `workflow_id` rather than only the answer is right too. "If operational evidence later shows a one to one relationship, that is an observation, not a constraint" is the sentence that stops someone encoding the correlation the first time they notice it.

Docs only, real checks green. Merging, and this is now the shape #566, #567 and #568 build against.

Thanks for opening it as a proposal and arguing the shape first. The five questions were the right five, and I would rather answer those than review an implementation of a guess.

PR #578, opened 2026-08-26
fix: cross-check intent_hash in agent manifest binding verification

### LINKED ISSUE #577: Agent Manifest binding does not cross-check intent_hash required by Agent Manifest spec §3.9
## cmcp-gateway version

git SHA `cf125a5` (`cf125a518e98bb23ed7e0be6abbeee208557167e`), forked from `agentrust-io/cmcp`, commit dated 2026-08-19.

## Python version

Python 3.13.1

## TEE provider

none (reproduced in software-only mode; the gap is in `cmcp_verify`'s comparison logic itself, independent of hardware attestation)

## Reproduction steps

1. Generate a signing keypair and create a signed Agent Manifest that declares an `intent` (e.g. `{"statement": "Move materials in Zone A only."}`).
2. Call `verify_agent_manifest_binding()` to obtain a `binding` from this manifest. At this point `binding.intent_hash` is not `None`; it is the correctly computed `sha256:...` digest of the declared `intent`.
3. Generate a TRACE Claim. Keep every other field of `gateway.agent_identity` (`manifest_id`, `agent_id`, `authenticated_subject`, `subject_source`, `issuer`, `issuer_key_id`, `policy_bundle_hash`, `tool_catalog_hash`) consistent with the `binding` from step 2, but set `intent_hash` to a different, well-formed but fabricated value (e.g. `"sha256:" + "f" * 64`).
4. Call `verify_trace_claim()` with this claim and the manifest from step 1 (plus `trusted_agent_manifest_keys`).
5. Inspect the returned `VerificationResult`'s `verified_fields` / `unverified_fields` / `failure_reason`.

A regression test reproducing this is added to `tests/unit/test_verify.py` (`test_agent_manifest_binding_intent_hash_mismatch_fails`, mirroring the existing `test_agent_manifest_binding_mismatch_fails`). Run with `pytest tests/unit/test_verify.py -v -k intent_hash`.

## Expected behavior

Like a mismatch on `agent_id` (see `test_agent_manifest_binding_mismatch_fails`), a mismatched `intent_hash` should be caught: `agent_manifest.binding` should land in `unverified_fields` with `failure_reason` set to `AGENT_MANIFEST_MISMATCH`.

## Actual behavior

The `intent_hash` mismatch is not caught. `agent_manifest.binding` lands in `verified_fields`. Overall `status` is `PARTIALLY_VERIFIED` with `failure_reason=None`; nothing in the result signals that `intent_hash` was fabricated.

Root cause: in `src/cmcp_verify/verify.py`, Step 5's `expected_identity` dict (around line 752) does not include `intent_hash`. `verify_agent_manifest_binding()` (`src/cmcp_runtime/agent_manifest.py`, lines 367-380) correctly computes `intent_hash` after signature verification and includes it on the returned `binding`, but `verify_trace_claim()`'s comparison loop (`mismatched = [key for key, expected in expected_identity.items() if agent_identity.get(key) != expected]`) only iterates over that dict's keys, so `intent_hash` is never compared regardless of its value.

Existing tests (`tests/unit/test_intent_binding.py`) cover computation, propagation into the emitted claim, and format validation of `intent_hash`, but not this cross-check at verification time.

## Why this matters

Of the 8 fields Step 5 currently compares (`manifest_id`, `agent_id`, `authenticated_subject`, `subject_source`, `issuer`, `issuer_key_id`, `policy_bundle_hash`, `tool_catalog_hash`), all describe *whose* agent this is and *under what configuration*. `intent_hash` is the only one that can express *for what declared purpose*, and none of the others substitute for it.

Separately, agent-manifest's spec is explicit (section 3.9.1) that nothing in this ecosystem measures whether an agent's actual actions were semantically consistent with its declared intent; that is deliberately out of scope. This makes the mechanical `intent_hash` equality check the only safeguard on the "declared purpose" axis at all. Compare with `policy_bundle_hash` or `tool_catalog_hash`, which overlap with independent runtime policy enforcement; intent has this one check and only this one, and it is the one currently missing.

```
VerificationResult(
    status=<VerificationStatus.PARTIALLY_VERIFIED: 'partially_verified'>,
    verified_fields=['schema', 'signature', 'policy_bundle.hash', 'tool_catalog.hash', 'agent_manifest.binding', 'attestation_freshness', 'audit_chain'],
    unverified_fields=['hardware_attestation'],
    failure_reason=None,
    attestation_age_seconds=0,
    is_attestation_fresh=True,
    details={
        'audit_chain_binding': 'software-only mode -- chain-root binding not applicable',
        'hardware_attestation': 'software-only mode - not hardware-backed',
    },
)
```

`hardware_attestation` in `unverified_fields` is an unrelated artifact of software-only mode, not a signal about the `intent_hash` mismatch.

None.

## Related

- #576 reports the same category of gap in the same verification path (`cMCP` ↔ Agent Manifest binding), for a different field: the runtime `enforcement_mode` is not cross-checked against `artifacts.policy_bundle.enforcement_mode` (Agent Manifest spec §6.2). That issue and this one both point at Step 5's field list being incomplete relative to what the spec expects to be bound.

yongikim:
## What

Add `intent_hash` to the field set that Step 5 (`verify_trace_claim`) cross-checks against the Agent Manifest binding, and add a regression test covering the mismatch case.

## Why

Closes #577. `expected_identity` in `src/cmcp_verify/verify.py` compares 8 fields between the Claim's `gateway.agent_identity` and the recomputed `AgentManifestBinding`, but omits `intent_hash`, even though `verify_agent_manifest_binding()` already computes it correctly on `binding.intent_hash`. A Claim can currently assert a fabricated `intent_hash` and still pass `agent_manifest.binding` verification.

## Security impact

Low current exploitability: the `intent` field is not yet read by any policy or tool-catalog enforcement path, so this does not bypass an active control. But it removes the only mechanical check on an agent's declared purpose (Agent Manifest spec §3.9), and the spec itself (§3.9.1) notes there is no semantic backstop for this anywhere else in the ecosystem. Does not touch `src/cmcp_gateway/audit/`, `src/cmcp_gateway/tee/`, or `src/cmcp_gateway/policy/`.

## Test plan

- [x] `pytest tests/unit/ -v` passes (adds `test_agent_manifest_binding_intent_hash_mismatch_fails` to `tests/unit/test_verify.py`, mirroring the existing `test_agent_manifest_binding_mismatch_fails`)
- [x] `ruff check src/cmcp_verify/verify.py tests/unit/test_verify.py` passes
- [x] `mypy src/cmcp_verify/verify.py` passes
- [x] `bandit -r src/ -c pyproject.toml` passes
- [ ] Manual test performed (describe steps below if applicable)

## DCO sign-off

### DIFF
--- a/src/cmcp_verify/verify.py
+++ b/src/cmcp_verify/verify.py
@@ -863,6 +863,7 @@ def verify_trace_claim(
                     "issuer_key_id": binding.issuer_key_id,
                     "policy_bundle_hash": binding.policy_bundle_hash,
                     "tool_catalog_hash": binding.tool_catalog_hash,
+                    "intent_hash": binding.intent_hash,
                 }
                 mismatched = [
                     key
--- a/tests/unit/test_verify.py
+++ b/tests/unit/test_verify.py
@@ -11,7 +11,11 @@ from datetime import UTC, datetime, timedelta
 from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
 from cryptography.hazmat.primitives.serialization import Encoding, PublicFormat
 
-from cmcp_runtime.agent_manifest import SIGNED_FIELDS, signing_pre_image
+from cmcp_runtime.agent_manifest import (
+    SIGNED_FIELDS,
+    signing_pre_image,
+    verify_agent_manifest_binding,
+)
 from cmcp_runtime.audit.chain import AuditChain
 from cmcp_runtime.audit.keys import SigningKey
 from cmcp_runtime.audit.trace_claim import (
@@ -142,7 +146,9 @@ def _b64url(data: bytes) -> str:
     return base64.urlsafe_b64encode(data).rstrip(b"=").decode()
 
 
-def _signed_manifest(priv: Ed25519PrivateKey, key_id: str) -> dict:
+def _signed_manifest(
+    priv: Ed25519PrivateKey, key_id: str, *, intent: dict | None = None
+) -> dict:
     manifest = {
         "@context": "https://agentmanifest.agentrust-io.com/v0.1/context.json",
         "@type": "AgentManifest",
@@ -159,6 +165,8 @@ def _signed_manifest(priv: Ed25519PrivateKey, key_id: str) -> dict:
         },
         "delegation_chain": [],
     }
+    if intent is not None:
+        manifest["intent"] = intent
     manifest["signature"] = {
         "algorithm": "Ed25519",
         "key_id": key_id,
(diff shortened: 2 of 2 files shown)


imran-siddique (approved):
Approving to release the hold gate. Three workflows were sitting in `action_required`, so cmcp's CI had not run on this at all: two checks reported where this repo normally runs twelve. Releasing them now and I will report before merging.

**Verified the defect at `main` rather than taking the description.** `expected_identity` at `verify.py:857` holds eight fields and omits `intent_hash`, while `binding.intent_hash` is computed at `agent_manifest.py:379` through `agent_manifest_sdk.intent_hash(manifest)`. So the value is derived correctly and then dropped on the floor at the comparison. A claim can assert a fabricated `intent_hash` and pass binding verification, exactly as you say.

This is the same defect class as cmcp#557 last week, where the chained TPM verifier parsed the signature scheme and digest and then passed only the bare signature. Computed, reported, not enforced. **Reported-but-unenforced is worse than absent**, because a field appearing in a verification result reads as checked to anyone downstream, and nobody re-derives it.

I also checked the `None` path, since `intent_hash` is typed `str | None` and an issuer may declare no intent. Both sides `None` compares equal and passes; a claim asserting an `intent_hash` against a manifest that declares none mismatches and fails. Fail-closed in the direction that matters, so the one-line addition needs no guard around it. Worth stating because a reviewer glancing at a nullable field in an equality check would reasonably ask.

**Your security framing is the right level of claim and I want to credit it.** "Low current exploitability: the `intent` field is not yet read by any policy or tool-catalog enforcement path, so this does not bypass an active control" is the sentence most reports would have left out, because it makes the finding sound smaller. It also happens to be the sentence that makes the rest of the report trustworthy. Naming what it does cost, the only mechanical check on an agent's declared purpose with no semantic backstop elsewhere, is the argument that actually carries it.

Filing #577 first and then the PR, rather than opening the PR alone, also means the defect is recorded independently of whether this particular fix is the one that lands.

CI report to follow.

imran-siddique:
CI report, as promised. Releasing the three held workflows took this from two checks to thirteen, all green: tests on 3.11, 3.12 and 3.13 across ubuntu and windows, CodeQL, governance and the Docker build. `gate` is our maintainer hold rather than a result.

Merging. Thanks for filing #577 alongside it.

PR #579, opened 2026-08-26
verify: enforce tool transcript hash binding

athena-kanellatou:
## What

Enforce the binding between `trace.tool_transcript.hash` and `gateway.audit_chain.tip` when verifying an exported audit bundle.

The verifier now normalizes the runtime's raw audit-chain tip representation before comparing it with the prefixed transcript hash, so valid claims continue to verify while semantically mismatched bindings fail closed.

## Why

`generate_trace_claim()` derives `trace.tool_transcript.hash` from the audit-chain tip, but the runtime stores `gateway.audit_chain.tip` as the raw digest while the transcript field is rendered as `sha256:<digest>`.

Without normalization, a verifier-side equality check would reject valid exported bundles purely because of representation differences.

This change adds explicit verification for the binding and a regression test covering a re-signed claim whose transcript hash is tampered while the claim signature remains valid.

## Testing

- Full CI passed on Python 3.11, 3.12, and 3.13
- Ubuntu and Windows matrix jobs passed
- Governance check passed
- Regression coverage added in `tests/unit/test_verify_command.py`

### DIFF
--- a/src/cmcp_verify/verify.py
+++ b/src/cmcp_verify/verify.py
@@ -658,6 +658,13 @@ def verify_audit_bundle(
 
     if claim_json is not None:
         chain = claim_json.get("gateway", {}).get("audit_chain", {})
+        transcript = claim_json.get("trace", {}).get("tool_transcript", {})
+        transcript_hash = transcript.get("hash")
+
+        if transcript_hash is not None and transcript_hash != chain.get("tip"):
+            failures.append(
+                "claim trace.tool_transcript.hash does not match gateway.audit_chain.tip"
+            )
         if chain.get("root") != entries[0].get("entry_hash"):
             failures.append("bundle root does not match claim gateway.audit_chain.root")
         if chain.get("tip") != entries[-1].get("entry_hash"):
--- a/src/cmcp_verify/verify.py
+++ b/src/cmcp_verify/verify.py
@@ -662,7 +662,14 @@ def verify_audit_bundle(
         transcript_hash = transcript.get("hash")
         chain_tip = chain.get("tip")
 
-        if isinstance(transcript_hash, str) and isinstance(chain_tip, str):
+        bundle_has_tool_calls = any(
+            entry.get("entry_type") == "tool_call" for entry in entries
+        )
+        if bundle_has_tool_calls and not isinstance(transcript_hash, str):
+            failures.append(
+                "claim trace.tool_transcript.hash is missing for a bundle with tool calls"
+            )
+        elif isinstance(transcript_hash, str) and isinstance(chain_tip, str):
             normalized_tip = (
                 chain_tip
                 if chain_tip.startswith(("sha256:", "sha384:"))
--- a/src/cmcp_verify/verify.py
+++ b/src/cmcp_verify/verify.py
@@ -660,11 +660,18 @@ def verify_audit_bundle(
         chain = claim_json.get("gateway", {}).get("audit_chain", {})
         transcript = claim_json.get("trace", {}).get("tool_transcript", {})
         transcript_hash = transcript.get("hash")
+        chain_tip = chain.get("tip")
 
-        if transcript_hash is not None and transcript_hash != chain.get("tip"):
-            failures.append(
-                "claim trace.tool_transcript.hash does not match gateway.audit_chain.tip"
+        if isinstance(transcript_hash, str) and isinstance(chain_tip, str):
+            normalized_tip = (
+                chain_tip
+                if chain_tip.startswith(("sha256:", "sha384:"))
+                else f"sha256:{chain_tip}"
             )
+            if transcript_hash != normalized_tip:
+                failures.append(
+                    "claim trace.tool_transcript.hash does not match gateway.audit_chain.tip"
+                )
         if chain.get("root") != entries[0].get("entry_hash"):
             failures.append("bundle root does not match claim gateway.audit_chain.root")
         if chain.get("tip") != entries[-1].get("entry_hash"):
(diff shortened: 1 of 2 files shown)


imran-siddique (dismissed):
Approving to release the hold gate, and releasing the workflows. Not merging yet: three runs were sitting in `action_required`, so our CI had not run on this at all. Your "full CI passed on 3.11, 3.12 and 3.13" is your fork's; ours reported two checks where cmcp normally runs twelve. Report to follow.

**The gap is real and I confirmed it.** `tool_transcript` appears zero times in `src/cmcp_verify/verify.py` on `main`. The verifier checks `chain.root` and `chain.tip` against the bundle entries, so the chain is bound to the bundle, and nothing bound the TRACE claim's transcript hash to that chain. A claim could name any transcript.

**Your regression test picks the right adversary.** Tampering the transcript hash on a re-signed claim whose signature remains valid is the case that matters, because a signature does not help when the attacker controls the key that made it. A test that tampered the field and left the signature broken would have proved only that signature verification works.

The normalization is also the right shape rather than a workaround. The runtime stores the tip as a raw digest and the transcript field is rendered `sha256:<digest>`, so an equality check without it would reject valid bundles, and "reject everything" is not fail-closed, it is broken. Accepting an already-prefixed `sha256:` or `sha384:` tip unchanged means a future digest change does not silently double-prefix.

## One question before I merge

The check sits inside:

```python
if isinstance(transcript_hash, str) and isinstance(chain_tip, str):
```

so it skips when either is absent. For a claim with no transcript that is correct: `trace_claim.py:222` types `tool_transcript` as `ToolTranscriptOut | None = None`, so absence is legitimate and failing there would reject valid bundles.

What I have not traced is whether the same branch is reachable by *removing* the field from a claim that had one. If a stripped `tool_transcript` still verifies its signature, then an attacker who can re-sign, which is the adversary your own test assumes, can delete the binding rather than tamper with it and take the skip path instead of the failure path.

Two things would settle it, and you are better placed than me to say which applies:

1. If removal breaks signature verification, the branch is unreachable for a tampered claim and this is correct as written. Worth a one-line comment saying so, since the next reader will ask the same question.
2. If removal survives, the skip needs to be conditional on something other than the field's presence, for instance requiring the transcript when the bundle carries entries.

Not asking you to widen the change. If it is case 1, say so and I will merge.

athena-kanellatou:
Thank you,I think this is case 2 under the same adversary model as the regression test.
tool_transcript is optional in the claim model, so if the claim is modified to remove it and then re-signed with the same valid key, signature verification itself should not make that branch unreachable. 

The signature authenticates the modified claim, it does not establish that an optional field was previously present.
That means the current presence-based guard can potentially turn deletion of the binding into a skip path.

I’ll add a focused regression case for a re-signed claim with tool_transcript removed and tighten the condition so that a bundle which carries transcript/audit-chain evidence cannot silently lose that binding, while preserving legitimately transcript-less claims.

Thanks for catching this before merge.

athena-kanellatou:
I’ve pushed the focused removal regression case on 79277e5. The upstream workflows are back behind the maintainer approval gate after the new commit.

athena-kanellatou:
I’ve pushed the completed removal-case fix in 195f6f8.

The regression is now a standalone test and confirms that a re-signed claim with tool_transcript removed still passes signature verification but fails audit-bundle verification when the bundle contains tool-call evidence. Legitimately transcript-less bundles without tool calls remain unaffected.

Local verification: 11 tests passed in tests/unit/test_verify_command.py. The upstream workflows are awaiting maintainer approval.

imran-siddique (approved):
You worked out the answer and it was the bad case. Your sentence is the one that settles it:

> The signature authenticates the modified claim, it does not establish that an optional field was previously present.

That is exactly right and it is the thing that makes optional fields dangerous in signed objects generally. Under an adversary who can re-sign, which is the adversary your own regression test already assumed, absence is indistinguishable from never-was, so a presence-based guard turns deletion into a skip rather than a failure.

**The fix keys on the right thing.** Verified at the head:

```python
bundle_has_tool_calls = any(entry.get("entry_type") == "tool_call" for entry in entries)
if bundle_has_tool_calls and not isinstance(transcript_hash, str):
    failures.append("claim trace.tool_transcript.hash is missing for a bundle with tool calls")
elif isinstance(transcript_hash, str) and isinstance(chain_tip, str):
    ...
```

The condition now depends on what the bundle contains rather than on what the claim declares. An attacker who deletes the transcript hash to escape the comparison would also have to delete the tool-call entries from the bundle to escape this, and those are what they are trying to keep. That is the property that makes it hold, and it is why "require the field when tool calls exist" is stronger than "require the field".

Legitimately transcript-less bundles with no tool calls still pass, so the change does not reject valid evidence. Getting both directions right is the part that usually needs a second round.

**Making the removal case a standalone regression asserting both halves is the right construction.** A re-signed claim with the transcript removed passes signature verification and fails audit-bundle verification. A test that only asserted the failure would not have shown that signature verification is not the control here, which is the whole point.

Two smaller things worth saying. Reporting your local count as 11 tests in one file, rather than a whole-suite number, is more useful than the larger figure. And flagging each time the push put the workflows back behind the approval gate saved me working out why the checks had reset.

Releasing the held runs now. CI report before I merge.

imran-siddique:
CI report, as promised. Releasing the held workflows took this from one check to twelve, all green. `gate` is our maintainer hold rather than a result.

Merging. The verifier now binds the TRACE transcript hash to the audit chain, and a claim that deletes the field cannot escape the check when the bundle carries tool calls.

Thanks for working out the deletion case rather than defending the original guard. "The signature authenticates the modified claim, it does not establish that an optional field was previously present" is the sentence I will be reusing about optional fields in signed objects.

PR #580, opened 2026-08-26
fix(ci): pin sbom.yml actions to SHA and scope permissions to the job

rajnisht7:
## What

This PR scopes permission of sbom.yml to the job and also updates the actions to SHA.

Part of the CI hardening pass in agentrust-io/.github#30.

## Why

`sbom.yml` ran `anchore/sbom-action@v0` (a floating tag) in a job that also holds `id-token: write`. A floating tag means whoever controls that tag controls what runs in a job that can mint an OIDC token and write to the repo and this is the single worst case flagged in the audit. Pinned it, along with `actions/checkout` and `actions/upload-artifact`, to exact commit SHAs with version comments.

Also moved `permissions: { contents: write, id-token: write }` from the workflow level down to the `sbom` job specifically, so a future job added to this file doesn't inherit them by default.

## Security impact

None

## Test plan

- [ ] `pytest` passes
- [ ] `ruff check` passes
- [ ] `mypy` passes

It is CI-config change:
- Parsed the file with `yaml.safe_load()` to confirm valid YAML syntax
- Confirmed each pinned SHA resolves to the exact version tag named in its comment, via `git ls-remote --tags` (e.g. `anchore/sbom-action`'s SHA matches its `v0.24.0` tag)
- Diffed against the original to confirm no step, input, or job logic changed and only the action version references and where the `permissions` block sits

## DCO sign-off

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -9,22 +9,21 @@ on:
 env:
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: "true"
 
-permissions:
-  contents: write
-  id-token: write
-
 jobs:
   sbom:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Generate SBOM
-        uses: anchore/sbom-action@v0
+        uses: anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
         with:
           format: spdx-json
           output-file: sbom.spdx.json
       - name: Upload SBOM artifact
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: sbom
           path: sbom.spdx.json


imran-siddique (approved):
Verified every SHA at source rather than trusting the version comments, since a wrong SHA in a pinning PR is worse than no pin: it looks pinned, passes review, and nobody re-checks it.

| pin | tag | result |
|---|---|---|
| `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1` | v7.0.1 | exact |
| `actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065` | v5.6.0 | exact |
| `actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a` | v7.0.1 | exact |
| `anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610` | v0.24.0 | exact |

Annotated tags dereferenced to their commits where needed, which is the step that makes a codeql-style pin look wrong when it is right.

**The permissions change is the part I would have asked for if it were missing.** Moving `contents: write` and `id-token: write` from workflow scope to job scope narrows them to the job that needs them, so every other job in that file drops to the default. `id-token: write` in particular is the one worth scoping tightly, since it mints OIDC tokens. Doing that in the same PR as the pinning is coherent rather than scope creep: both are about what a compromised step can reach.

**Keeping the `# vX.Y.Z` comment beside each SHA** is what makes this maintainable. A bare SHA is unreadable and gets bumped by someone pasting a newer one without checking what it is. Saying explicitly in the `.github` note that the SHA and the comment must be updated together is the instruction that keeps them from drifting apart.

Replicating the identical change across ca2a, cmcp and trace-spec as separate PRs rather than one is right for repos that do not share a workflow file. They are byte-identical, which I checked.

Merged.

PR #584, opened 2026-08-26
fix(agent-manifest): wire runtime enforcement_mode into manifest verification

### LINKED ISSUE #576: Agent Manifest binding does not cross-check cMCP enforcement_mode required by §6.2
## Problem

Agent Manifest v0.2 §6.2 requires the cMCP attestation `enforcement_mode` to match the Agent Manifest `artifacts.policy_bundle.enforcement_mode`, with a conflict treated as an attestation failure.

The current cMCP → Agent Manifest verification path appears to bind runtime policy and tool catalog hashes through `agent_manifest.verify_manifest()`, but does not pass or compare the runtime `enforcement_mode`.

This looks like the unfinished follow-up already noted in #342:

> Extend to cover enforcement_mode attestation binding (Claim 1, Property 3)

That issue was closed after the hash-binding experiment shipped, but the enforcement-mode follow-up itself does not appear to have been implemented.

## Reproduction / current path

The current Agent Manifest integration uses `VerificationContext` with values such as:

* `policy_bundle_hash`
* `tool_catalog_hash`
* trusted signing keys

but there is no corresponding runtime enforcement-mode value passed into the verifier.

At the same time, cMCP has runtime modes:

* `enforcing`
* `advisory`
* `silent`

while Agent Manifest defines:

* `enforce`
* `advisory`
* `audit-only`

and §6.2 requires the two sides to match.

Example mismatch:

```text
cMCP runtime enforcement_mode = advisory
Agent Manifest policy_bundle.enforcement_mode = enforce
```

The current binding path appears to still be able to verify the policy hash/catalog hash without explicitly rejecting this mode mismatch.

## Expected behavior

The cMCP → Agent Manifest verification path should cross-check the runtime/attested enforcement mode against `artifacts.policy_bundle.enforcement_mode`.

A mismatch should fail verification / attestation rather than returning a valid binding.

## Actual behavior

The canonical Agent Manifest verification integration checks the policy bundle hash and other runtime bindings, but does not appear to perform an enforcement-mode comparison.

## Impact

This is narrower than a policy-bundle hash bypass: the policy bundle itself can still match exactly.

The gap is that the same approved policy bundle could be present while the runtime is operating in a different enforcement mode than the manifest declares.

## Suggested narrow fix

One possible implementation would be:

1. Define the cMCP ↔ Agent Manifest mode mapping explicitly:

   * `enforcing` → `enforce`
   * `advisory` → `advisory`
   * `silent` → `audit-only` (if this is the intended semantic mapping)
2. Pass the normalized runtime mode into the Agent Manifest verification path.
3. Reject a mismatch with a non-VALID result.
4. Add a negative test for:

   * manifest `enforce`
   * runtime `advisory`
   * expected result: mismatch / attestation failure

Related:

* #342
* #323
* agentrust-io/agent-manifest#176
* agentrust-io/agent-manifest#178

rajnisht7:
## What

Adds a runtime enforcement-mode cross-check to Agent Manifest binding verification. Previously only the policy_bundle hash was checked against the manifest in the runtime's actual enforcement mode (enforce / advisory / audit-only) and was never verified against what the manifest declares, so a runtime could silently run in a weaker mode than its manifest requires while still passing verification.

kindly merge this after https://github.com/agentrust-io/agent-manifest/pull/345
## Why

Fixes: #576

A manifest's policy_bundle can declare enforcement_mode: "enforce", but cMCP never checked the running instance was actually attested as being in that mode. Same approved policy bundle, same hash match, but the runtime could be configured for advisory (warn-only, doesn't block) or silent, and verify_agent_manifest_binding would still report success. This closes that gap: cMCP now passes its own config.attestation.enforcement_mode into the verification call (via a small mapping, since cMCP's own vocabulary enforcing / advisory / silent that doesn't match the Agent Manifest spec's enforce / advisory / audit-only), and a mismatch (or an unattested mode when the manifest requires one) now fails closed instead of passing silently.

Companion fix in agentrust-io/agent-manifest (adds the enforcement_mode field and cross-check to VerificationContext / verify_manifest this PR is the cMCP-side consumer of that).

## Security impact

Yes. This touches capability/policy enforcement verification.

## Test plan

- [x] `pytest` passes
- [x] `ruff check` passes
- [x] `mypy` passes

Executed:

pytest tests/unit/test_agent_manifest.py tests/unit/test_agent_manifest_cose.py -q
ruff check src/cmcp_runtime/agent_manifest.py src/cmcp_runtime/startup.py tests/unit/test_agent_manifest.py tests/unit/test_agent_manifest_cose.py
mypy src/cmcp_runtime/agent_manifest.py src/cmcp_runtime/startup.py

## DCO sign-off


rajnisht7:
This PR depends upon agentrust-io/agent-manifest#345, tested locally against the patched SDK and confirmed CI turns fully green once that merges (verified the exact 2 failing tests pass, plus re-ran the full related test suite).

One more thing found while tracing this:` src/cmcp_verify/verify.py's` `verify_trace_claim()` also calls `verify_agent_manifest_binding()` ( the optional Agent Manifest binding cross-check) but doesn't pass enforcement_mode. It's safely wrapped in try/except ConfigError so nothing crashes, and no current test exercises a manifest with enforcement_mode declared through that path, so this PR's CI stays green but once agentrust-io/agent-manifest#345 ships, any manifest that declares enforcement_mode (which spec says is REQUIRED) will report agent_manifest.binding: unverified there, because there's currently no "claimed enforcement mode" recorded in AgentIdentityInfo/TRACE claims to check against. Happy to file a follow-up issue for that, or take a stab at it didn't want to fold it into this PR since it needs a claim-schema change (recording enforcement_mode at claim-creation time), which felt like separate scope.

imran-siddique:
One required release fix before approval: please change gent-manifest>=0.11.1 to gent-manifest>=0.11.2 in pyproject.toml. This PR passes nforcement_mode= to VerificationContext, but 0.11.1 has no such field; leaving the old floor permits a metadata-valid install that fails on the new verification path. 0.11.2 is now published and contains agentrust-io/agent-manifest#345. Once that one-line floor update lands, I’ll re-review the refreshed green head and merge.

imran-siddique:
Correction to the formatting in my previous comment: change the dependency requirement from agent-manifest>=0.11.1 to agent-manifest>=0.11.2 in pyproject.toml. Version 0.11.2 is now published. This floor is required because VerificationContext(enforcement_mode=...) is not supported by 0.11.1.

imran-siddique (approved):
Verified the enforcement-mode binding against agent-manifest 0.11.2. Local evidence: 1,492 passed, 8 skipped; Ruff clean; mypy clean across 65 source files. Dependency floor now matches the required SDK API and behavior.

Issue #147, opened 2026-06-06
HIGH: No request body size limit before JSON parsing — pre-auth OOM via oversized payload (INJECT-001/DOS-001)


imran-siddique:
`_handle_mcp()` calls `await request.body()` with no size cap before `json.loads()`. A client can send an arbitrarily large HTTP body, causing the gateway to buffer the full payload in memory before detecting malformed JSON. This is pre-authentication: the size check runs before any auth check. Also, `json.loads()` has no depth or key-count limit — a deeply nested JSON object can cause RecursionError or extreme CPU usage. Fix: add a `MAX_REQUEST_BODY = 1 MB` cap before reading the body.

PR #585, opened 2026-08-27
docs: settle MAX_REQUEST_BYTES provenance (#573)

### LINKED ISSUE #573: MAX_REQUEST_BYTES: spec says 10MB, both implementations enforce 1MB
`docs/spec/proxy-security.md:39` specs `MAX_REQUEST_BYTES = 10 * 1024 * 1024  # 10MB`.

Both implementations enforce 1MB:

- `scripts/mock_upstream.py`, `MAX_REQUEST_BYTES = 1_000_000`
- `src/cmcp_runtime/mcp/server.py`, `_DEFAULT_MAX_REQUEST_BYTES = 1_000_000`, used as the `max_request_bytes` default on `MCPServer.__init__`

An order of magnitude apart, and nothing in the repo records which one is intended. I raised this on #562 while implementing `MAX_STRING_LENGTH` and deliberately didn't pick a side, but #562 was closed as completed by #570, so the question isn't tracked anywhere now. Opening this so it isn't lost.

The two readings lead opposite ways:

- **1MB was a deliberate tightening**, and the spec is stale. Fix the document.
- **1MB is drift**, and implementations should be raised toward 10MB. Fix the code, and note that widening a DoS bound is not a change to make quietly.

I don't have evidence for either. The spec is Draft v0.1 and its git history doesn't say where 10MB came from.

Two things that depend on the answer, so it's worth settling rather than leaving:

`MAX_STRING_LENGTH` is now derived from whatever body cap is enforced (#570, and #572 makes the spec say so), which removes the dead-code hazard at either value. So this doesn't block that. But the derived cap moves with the body cap, and a jump from 1MB to 10MB moves the per-string cap from 500KB to 5MB, which is a real change in what the gateway accepts rather than a documentation cleanup.

`MAX_JSON_NESTING_DEPTH` and `MAX_PARSE_TIME_MS` are still unimplemented. Whoever implements them will size them against a body cap, and picking against the wrong one bakes the mismatch in further.

Happy to send whichever patch the answer implies. I don't think it's a contributor's call to make.

Yatsuiii:
## Summary

Records the maintainer decision from #573 in docs/spec/proxy-security.md.

1. Sets MAX_REQUEST_BYTES to 1_000_000 decimal.
2. Records DOS-001 issue #147 as the provenance.
3. Documents the derived 500,000 byte string limit.
4. States that MAX_JSON_NESTING_DEPTH and MAX_PARSE_TIME_MS must be evaluated against the 1MB cap.
5. Keeps the 10MB value as a separate resource and denial of service design decision.

No runtime code changed.

## Verification

The change is limited to docs/spec/proxy-security.md.

Closes #573.

### DIFF
--- a/docs/spec/proxy-security.md
+++ b/docs/spec/proxy-security.md
@@ -2,7 +2,7 @@
 
 ---
 Status: Draft v0.1
-Last updated: 2026-06-04
+Last updated: 2026-08-27
 Stability: Unstable , expect breaking changes before v1.0
 ---
 
@@ -35,7 +35,7 @@ All items must be satisfied before Phase 2 ships. Items are non-negotiable; a pa
 These constants are hard-coded in the proxy implementation. They are not configurable at runtime or via operator-supplied configuration. Making them configurable would allow an operator to raise limits and defeat the protection.
 
 ```python
-MAX_REQUEST_BYTES = 10 * 1024 * 1024  # 10MB
+MAX_REQUEST_BYTES = 1_000_000  # 1MB, decimal
 MAX_JSON_NESTING_DEPTH = 64
 MAX_PARSE_TIME_MS = 100
 MAX_STRING_LENGTH = MAX_REQUEST_BYTES // 2   # per string field, see invariant below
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):
Verified against both implementations and the DOS-001 provenance in #147/#573. The patch correctly documents the enforced decimal 1,000,000-byte cap and derived 500,000-byte string limit.

PR #586, opened 2026-08-27
fix(verify): bind call summary to audit evidence

### LINKED ISSUE #582: Audit-bundle verification does not bind call_summary to tool-call evidence
What happened

"verify_audit_bundle()" verifies the exported entry hash chain, the bundle signature, and the claim's "gateway.audit_chain" root, tip, and length. After #579 it also binds "trace.tool_transcript.hash" to the chain tip.

It does not, however, re-derive the audit-derived call metadata carried by the same claim:

- "trace.tool_transcript.call_count"
- "gateway.call_summary.tool_calls_total"
- "gateway.call_summary.tool_calls_allowed"
- "gateway.call_summary.tool_calls_denied"
- "gateway.call_summary.tool_calls_faulted"
- "gateway.call_summary.tools_invoked"

The runtime derives these fields from audit entries when closing a session, but the offline verifier currently accepts a validly re-signed claim whose summaries contradict the supplied, correctly signed audit bundle.

This is the verifier-side counterpart of the producer requirement in #49 that call-summary totals be derived from audit-chain entries.

Reproduction

Starting from a real claim and exported audit bundle containing one allowed call to "fixture-tool":

1. Change both "trace.tool_transcript.call_count" and "gateway.call_summary.tool_calls_total" to the same incorrect value.
2. Change the summary to report:
   - "tool_calls_allowed = 0"
   - "tool_calls_denied = 1"
   - "tools_invoked = ["substituted.tool"]"
3. Re-sign the modified claim with the original test signing key.
4. Run "cmcp verify CLAIM --audit-bundle BUNDLE".

Current result:

signature                PASS
audit_bundle             PASS

The signature authenticates the modified claim, but it does not establish that the summary agrees with the separately supplied audit evidence.

An existing conformance fixture also exposes the missing appraisal: its bundle records "robot.request_motion", while its helper-generated claim reports "tools_invoked = ["tool.a"]"; the bundle currently verifies.

Expected behavior

When a claim and audit bundle are supplied together, the verifier should derive from entries whose "entry_type == "tool_call"":

- total call count,
- allowed count from "policy_decision == "allow"",
- denied count from "policy_decision in ("deny", "advisory_deny")",
- faulted count from "policy_decision == "fault"",
- invoked tools from the unique non-null "tool_name" values.

A mismatch with the corresponding claim fields should fail audit-bundle verification.

"session_max_sensitivity", "call_graph_summary", and "len(tool_transcript.entries)" should remain out of scope because they are not always reconstructible from the exported tool-call entries alone.

Environment

- cMCP git SHA: "b61a467" (current "main", merged #579)
- Python: 3.12.13
- TEE provider: software-only test fixture

athena-kanellatou:
What

Bind the claim’s audit-derived call metadata to the supplied audit bundle during offline verification.

"verify_audit_bundle()" now re-derives the tool-call count, allowed/denied/faulted counts, and unique invoked tools from "tool_call" entries, then compares them with "trace.tool_transcript.call_count" and the five corresponding "gateway.call_summary" fields.

Why

A validly re-signed claim could previously carry call summaries that contradicted the separately supplied, correctly signed audit evidence while both signature and audit-bundle verification passed.

Closes #582.

Security impact

This strengthens audit-chain integrity by making contradictory call metadata a fatal audit-bundle verification failure, even when the modified claim has a valid signature. The verifier uses the same derivation semantics as the runtime producer:

- total: entries with "entry_type == "tool_call""
- allowed: "policy_decision == "allow""
- denied: "policy_decision in ("deny", "advisory_deny")"
- faulted: "policy_decision == "fault""
- invoked tools: sorted unique non-null "tool_name" values

Fields that cannot always be reconstructed from exported tool-call entries remain out of scope.

Test plan

- [x] Targeted "pytest" verification tests pass ("59 passed")
- [x] "ruff check" passes for the modified verifier and test files
- [ ] "mypy" passes
- [ ] Manual test performed

Regression coverage independently mutates and re-signs each of the six bound fields, confirms claim signature verification still passes, and confirms audit-bundle verification fails with the corresponding mismatch.

DCO sign-off

### DIFF
--- a/src/cmcp_verify/verify.py
+++ b/src/cmcp_verify/verify.py
@@ -662,9 +662,8 @@ def verify_audit_bundle(
         transcript_hash = transcript.get("hash")
         chain_tip = chain.get("tip")
 
-        bundle_has_tool_calls = any(
-            entry.get("entry_type") == "tool_call" for entry in entries
-        )
+        tool_calls = [entry for entry in entries if entry.get("entry_type") == "tool_call"]
+        bundle_has_tool_calls = bool(tool_calls)
         if bundle_has_tool_calls and not isinstance(transcript_hash, str):
             failures.append(
                 "claim trace.tool_transcript.hash is missing for a bundle with tool calls"
@@ -679,6 +678,39 @@ def verify_audit_bundle(
                 failures.append(
                     "claim trace.tool_transcript.hash does not match gateway.audit_chain.tip"
                 )
+
+        call_summary = claim_json.get("gateway", {}).get("call_summary", {})
+        expected_call_summary = {
+            "tool_calls_total": len(tool_calls),
+            "tool_calls_allowed": sum(
+                1 for entry in tool_calls if entry.get("policy_decision") == "allow"
+            ),
+            "tool_calls_denied": sum(
+                1
+                for entry in tool_calls
+                if entry.get("policy_decision") in ("deny", "advisory_deny")
+            ),
+            "tool_calls_faulted": sum(
+                1 for entry in tool_calls if entry.get("policy_decision") == "fault"
+            ),
+            "tools_invoked": sorted(
+                {
+                    entry["tool_name"]
+                    for entry in tool_calls
+                    if entry.get("tool_name") is not None
+                }
+            ),
+        }
+        if transcript.get("call_count") != len(tool_calls):
+            failures.append(
+                "claim trace.tool_transcript.call_count does not match audit bundle tool calls"
+            )
+        for field_name, expected_value in expected_call_summary.items():
+            if call_summary.get(field_name) != expected_value:
+                failures.append(
+                    f"claim gateway.call_summary.{field_name} does not match audit bundle tool calls"
+                )
+
         if chain.get("root") != entries[0].get("entry_hash"):
             failures.append("bundle root does not match claim gateway.audit_chain.root")
         if chain.get("tip") != entries[-1].get("entry_hash"):
--- a/src/cmcp_verify/verify.py
+++ b/src/cmcp_verify/verify.py
@@ -701,12 +701,25 @@ def verify_audit_bundle(
                 }
             ),
         }
-        if transcript.get("call_count") != len(tool_calls):
+        actual_call_count = transcript.get("call_count")
+        if (
+            not isinstance(actual_call_count, int)
+            or isinstance(actual_call_count, bool)
+            or actual_call_count != len(tool_calls)
+        ):
             failures.append(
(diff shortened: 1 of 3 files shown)


athena-kanellatou:
@Yatsuiii The implementation is ready and CI is green. I’d appreciate your review when you have a chance, as you offered on #582. Thank you!

Yatsuiii:
Reviewed at `a211ec2`, as offered on #582.

The verifier's derivation matches the producer's in `src/cmcp_runtime/session/manager.py:265-279` field for field: same `entry_type == "tool_call"` filter, same `allow` / `("deny", "advisory_deny")` / `fault` predicates, same `sorted()` over the non-null `tool_name` set. The producer sorts too, so binding to sorted order will not reject a valid claim.

Ran the checkbox you left unticked: `mypy src` gives `Success: no issues found in 64 source files`. `tests/unit/test_verify_command.py` and `tests/conformance/test_audit_conformance.py` give `53 passed`.

Two things I checked that are fine. Empty `entries` cannot reach the new block, `verify.py:564` returns first. A claim with no `gateway.call_summary` fails closed, since `.get(field)` returns `None` against an integer.

Two findings, neither blocking.

`_summary(chain)` now derives the five fields with the same rule the verifier uses. That fixes the hardcoded `tools_invoked=["tool.a"]` that contradicted its own bundle, but the fixture can no longer disagree with the verifier, so a wrong derivation would pass on both sides. The six mutation cases pin the negative direction. The positive direction is now checked against a value the test computed itself. Literal expected values matching the fixture's chain would keep that pin.

Booleans satisfy the integer comparisons. A claim carrying `"tool_calls_total": true` passes against a real count of 1, because `True != 1` is `False`. Same for `false` against zero. The count is not misreported so I would not hold the PR for it, but `isinstance(v, int) and not isinstance(v, bool)` closes it. Same class as the `slsa_level` boolean in trace-tests#85.

Binding looks correct and each of the six fields has its own regression.

Yatsuiii:
Both addressed. The integer guard reads correctly, and pinning in the unit test rather than the fixture is the better call. Nothing further from me.

athena-kanellatou:
Thanks for the thorough review and confirmation!

imran-siddique (approved):
Reviewed at 46dee44. The verifier derives all six reconstructable fields with the same predicates and ordering as the runtime producer, rejects bool-as-int values, and the regression suite independently re-signs each contradictory claim so signature PASS / audit bundle FAIL is causal evidence. Empty bundles are already rejected before this block; no unresolved review threads remain; all six Linux/Windows CI matrix jobs, CodeQL, and governance passed. No blocking findings.

PR #589, opened 2026-08-28
fix(server): tools/call 500s on non-string name, silently accepts non-object arguments

rajnisht7:
## What

POST /mcp tools/call crashed with a 500 on a non-string name field, and silently accepted a non-object arguments field (bypassing the #518/#562 shape-validation gate). Both now return a proper -32602 Invalid params JSON-RPC error, matching how the adjacent _cmcp field and the shape-check already handle malformed input.

## Why

_handle_tool_call read tool_name: str = params.get("name", "").lower() and arguments: dict[str, Any] = params.get("arguments", {}). The .get(field, default) default only covers a genuinely absent field, a name that is present but not a string (an int, a list, a bool, null) reached .lower() and raised an unhandled AttributeError, caught only by the outermost _unhandled_error_handler and logged as UNHANDLED_EXCEPTION/INTERNAL_ERROR for what is ordinary client input validation. arguments had the matching gap the other way: _arg_shape_violation only recognizes dict, list and str, so a scalar arguments value silently passed the shape check and reached call_tool with a shape its own type annotation says cannot occur.

Both fields are exactly as caller-controlled as _cmcp a few lines below, already guarded with "A malformed _cmcp (string, list, number) must not 500 the call path" name and arguments were the two places that same reasoning wasn't applied.

## Security impact

Low-to-moderate. Neither bug crashes the server process (Starlette's outer exception handler catches the name case cleanly) or leaks internal state.

## Test plan

- [x] `pytest` passes
- [x] `ruff check` passes
- [x] `mypy` passes

## DCO sign-off


imran-siddique (approved):
Both holes are real and the fix is right. `params.get("name", "")` defends only against absence, so a `name` that is present but an int, list or bool reached `.lower()` and surfaced as `UNHANDLED_EXCEPTION / INTERNAL_ERROR` for what is ordinary client input validation. `arguments` had the mirror gap: `_arg_shape_violation` recognises dict, list and str, so a scalar passed the #518/#562 gate and reached `call_tool` with a shape its own annotation excludes.

I checked the pattern you say you are matching. The existing `-32602` site on main returns `status_code=400`, so your two guards are consistent with it.

**One correction to the framing, since the body becomes the commit message.** These do not match how `_cmcp` handles malformed input. `_cmcp` at server.py:577-580 *coerces*:

```python
cmcp_params = params.get("_cmcp")
if not isinstance(cmcp_params, dict):
    cmcp_params = {}
```

Yours *reject*. That is the better choice and I want it kept: `_cmcp` is optional metadata, so degrading to empty loses nothing, whereas `name` and `arguments` are the request, and silently proceeding with a malformed one is how a caller gets a confident answer to a question it did not ask. But the reasoning is "stricter than `_cmcp`, deliberately", not "same as `_cmcp`". Worth saying that way if you touch the body.

Approving and merging.

PR #590, opened 2026-08-28
fix(cmcp-verify): cross-check enforcement_mode in TRACE claim re-verification

rajnisht7:
## What

Extends the enforcement_mode cross-check from #584 (startup-time binding) to claim-time re-verification: `verify_trace_claim`'s optional Agent Manifest binding check now also cross-checks the enforcement mode recorded at claim-creation time against a fresh re-verification of the manifest.

## Why

Found while verifying #584 would go fully green: `verify_trace_claim` (src/cmcp_verify/verify.py) calls `verify_agent_manifest_binding` too, but never passed `enforcement_mode`. It's wrapped in try/except so it doesn't crash, and no current test exercises a manifest that declares enforcement_mode` through this specific path, so #584 stays green either way but once agent-manifest ships the enforcement_mode check, any manifest built to spec (enforcement_mode is REQUIRED per 6.2) would start silently reporting `agent_manifest.binding: unverified` here, with nothing actually wrong. This closes that gap before it can surface: `AgentIdentityInfo`/`AgentIdentityOut` now carry the enforcement mode
recorded at claim-creation time, and `verify_trace_claim` cross-checks it the same way it already does for `intent_hash`, `policy_bundle_hash`, etc.

## Security impact

Yes. Same class of change as #584/#576, applied to the claim re-verification path instead of the startup path. Before this change, a claim's recorded enforcement mode was never checked against anything an attacker who could modify a claim or a bug that mis-recorded it) could assert any enforcement mode with no cross-check catching it. After this change, the claimed mode must match what a fresh, cryptographic re-verification of the manifest actually attests, or the binding is reported unverified. Also hardens against a garbled/invalid claimed value (`ValueError` from an unrecognized `enforcement_mode` string) now fails closed via the same except path as every other binding failure, instead of raising unhandled.

## Test plan

- [x] `pytest` passes
- [x] `ruff check` passes
- [x] `mypy` passes

## DCO sign-off

- [ ] I certify that I wrote or have the right to submit this contribution, and I agree to the

### DIFF
--- a/src/cmcp_runtime/audit/trace_claim.py
+++ b/src/cmcp_runtime/audit/trace_claim.py
@@ -111,6 +111,7 @@ class AgentIdentityInfo:
     # subject_source is a non-live-authenticated source (config, manifest-dev) - see
     # AgentIdentityOut for the full contract.
     agent_key_thumbprint: str | None = None
+    enforcement_mode: str | None = None
 
 
 # ── Pydantic output models ─────────────────────────────────────────────────────
@@ -177,6 +178,7 @@ class AgentIdentityOut(BaseModel):
     tool_catalog_hash: str
     intent_hash: Annotated[str, Field(pattern=r"^sha256:[0-9a-f]{64}$")] | None = None
     agent_key_thumbprint: Annotated[str, Field(pattern=r"^sha256:[0-9a-f]{64}$")] | None = None
+    enforcement_mode: str | None = None
 
 
 class ToolTranscriptEntry(BaseModel):
@@ -508,6 +510,7 @@ def generate_trace_claim(
                 tool_catalog_hash=agent_identity.tool_catalog_hash,
                 intent_hash=agent_identity.intent_hash,
                 agent_key_thumbprint=agent_identity.agent_key_thumbprint,
+                enforcement_mode=agent_identity.enforcement_mode,
             )
             if agent_identity is not None
             else None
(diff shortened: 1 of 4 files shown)


athena-kanellatou:
One edge case I don't see covered yet: should we add a negative test for the asymmetric absence case where the Agent Manifest declares "enforcement_mode", but "gateway.agent_identity.enforcement_mode" is missing from the claim?

The current backward-compatibility test covers both sides omitting the field, which makes sense for older manifests. But once the manifest declares the §6.2 binding, should claim-time re-verification treat a missing claimed mode as a binding failure rather than passing "None" through as absence?

imran-siddique (approved):
Correct, and it closes the half #584 left open. `verify_trace_claim` was calling `verify_agent_manifest_binding` without `enforcement_mode`, so once a manifest declares one, the claim path would have reported `agent_manifest.binding: unverified` with nothing actually wrong. Catching that while verifying #584 rather than after a user hit it is the right instinct.

I checked the comparison for the trap this shape usually has: `binding.enforcement_mode` is an `EnforcementMode` while the claim carries a string, which would mismatch on every record. `EnforcementMode` is a `StrEnum`, so they compare equal. Fine.

**One non-blocking point on the except.** `ConfigError` derives from `CMCPError`, not from `ValueError`, so `except (ConfigError, ValueError)` is a genuine widening rather than a restatement. The `ValueError` you want is from `EnforcementMode(enforcement_mode_raw)` on an unrecognised string, and that is the right thing to fail closed on. But the same clause now also swallows any incidental `ValueError` raised inside `verify_agent_manifest_binding` and reports it as `unverified`, which is exactly the shape where a real bug in binding verification becomes invisible: the operator sees "unverified", assumes a mismatch, and never learns something threw. Constructing the enum in its own small `try` would keep the fail-closed behaviour without extending the blast radius. Not worth blocking on, worth doing next time you are in this file.

Approving and merging.

PR #596, opened 2026-08-30
fix(verify): preserve structural claim failure

### LINKED ISSUE #592: verify_trace_claim can raise TypeError after schema failure on malformed nested claim shapes
## Current behavior

`verify_trace_claim()` validates the claim with `RuntimeClaim.model_validate()` first, but a schema failure is recorded and verification continues into the signature/key-binding steps.

At current `main` (`2c0a601805bfb31fddc49db32ae122bae8e2251e`), `_verify_signature()` begins with:

```python
try:
    x_b64: str = claim["trace"]["cnf"]["jwk"]["x"]
    padding = 4 - (len(x_b64) % 4)
    ...
except (KeyError, ValueError) as exc:
    return False, ...
```

That guard does not cover malformed JSON shapes. Examples include:

- `trace` is a string/list instead of an object -> `TypeError` while indexing;
- `trace.cnf` or `trace.cnf.jwk` is a scalar -> `TypeError`;
- `trace.cnf.jwk.x` exists but is a non-string scalar -> `TypeError` from `len()`.

Because `verify_trace_claim()` continues after `_validate_schema()` has already classified the record as malformed, those exceptions can escape instead of producing a `VerificationResult` with `CLAIM_MALFORMED` / signature unverified.

The same structural assumption appears in later chained `.get()` reads such as:

```python
claim_json.get("trace", {}).get("cnf", {}).get("jwk", {}).get("x", "")
```

which are safe only if every intermediate value is an object.

## Expected behavior

Malformed externally supplied claim structure should never terminate the verifier. Once schema validation has failed, later checks should either:

1. not execute when they require schema-established structure; or
2. independently parse defensively and return an unverified/failure result.

The important invariant is that a malformed claim produces a complete `VerificationResult`, not an uncaught Python exception.

## Suggested narrow direction

Prefer making schema validity a prerequisite for structure-dependent verification steps rather than widening every local `except` clause. That keeps one structural authority instead of duplicating a partial shape validator through the verifier.

If maintainers want best-effort secondary findings even after schema failure, introduce an object-safe accessor/helper and test every nested discriminator path explicitly.

## Regression matrix

At minimum:

- `trace` = string/list/null;
- `trace.cnf` = string/list/bool;
- `trace.cnf.jwk` = string/list;
- `trace.cnf.jwk.x` = integer/list;
- malformed `gateway`, `gateway.audit_chain`, and `gateway.attestation_evidence` intermediates;
- valid claim control unchanged.

Each malformed vector should return a `VerificationResult` and never raise.

No schema or wire-format change is required; this is verifier exception-safety and failure-path consistency.

altrudev:
Closes #592.

## Maintainer ruling and correction

Reconstructed on current upstream `main` after the review of `141c5b2`. The strict schema prerequisite is preserved: any `RuntimeClaim` validation failure returns `CLAIM_MALFORMED` before signature, key-binding or other downstream interpretation. No general ordering between signature and key-binding failures is introduced.

The intermediate compatibility workaround at `877c506` was too broad and has been removed. The original strict structural boundary and its 15 regression tests are preserved byte-for-byte. The positive-path TPM fixtures now use the valid, signed `gateway.attestation_evidence` envelope rather than placing evidence in schema-closed `trace.runtime`.

## Fixture and representation corrections

- TPM raw evidence is supplied through the producer's `AttestationReportInfo` and included in the final envelope signature.
- The separately supplied EK certificate chain is passed through the producer fixture and signed with the rest of the claim, preserving the intended EK-chain verification assertion.
- A malformed measurement is now expected to stop at the schema boundary, rather than reaching the TPM parser.
- The cross-platform evidence test retains positive coverage for supported Azure CVM, AMD SEV-SNP and Intel TDX platforms. The old OPAQUE positive case used a platform value not permitted by TRACE 0.10.0. It is now an explicit refusal regression, along with legacy unsupported aliases; no schema expansion or alias promotion is proposed.

## Executable evidence

Exact candidate `6ffad28da75560dfba5a0c242839403f871e9c53`, parent `cc27f91ec9b2d623dc633e24f1c368fb20f7566f` (current upstream main at reconstruction). One DCO-signed commit, five changed files, no unrelated source changes.

- `ruff check src/ tests/`: PASS.
- `mypy src/cmcp_runtime/`: PASS, 54 source files.
- `mypy src/cmcp_verify/verify.py`: PASS.
- Focused structural/TPM/cross-platform matrix: 64 passed.
- Full `python -m pytest -q`: 1,671 passed, 14 skipped.
- CI-scoped unit/conformance/integration suite with coverage: 1,629 passed, 6 skipped; coverage 87.56%, above the required 70%.
- Original strict schema implementation and all 15 new regression tests were independently compared against `141c5b2` and found byte-for-byte identical.
- The earlier 25-failure candidate and its 5 TPM regressions are superseded; the current candidate has no observed test failures. Hosted checks on the replacement head were not yet claimed at the time of this description.

## Scope and lineage

No schema, wire-format, cryptographic primitive, or general failure-precedence change. The old draft history is replaced by a single signed-off commit on current main, preserving original implementation credit and making the DCO history complete.

AI-assistance disclosure: ChatGPT assisted with source triage, boundary analysis, implementation and fixture correction, differential testing, clean-history reconstruction, and exact-diff review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/src/cmcp_verify/verify.py
+++ b/src/cmcp_verify/verify.py
@@ -531,13 +531,29 @@ def _coerce_measurement_digest(value: str | bytes) -> bytes | None:
         return None
 
 
-def _validate_schema(claim: dict[str, Any]) -> tuple[bool, str | None]:
+def _validation_error_path(exc: ValidationError) -> str:
+    """Return the most specific non-missing schema location for diagnostics."""
+    errors = exc.errors()
+    for error in errors:
+        if error["type"] == "missing":
+            continue
+        loc = error["loc"]
+        if loc:
+            return ".".join(str(part) for part in loc)
+    for error in errors:
+        loc = error["loc"]
+        if loc:
+            return ".".join(str(part) for part in loc)
+    return "claim"
+
+
+def _validate_schema(claim: dict[str, Any]) -> tuple[bool, str | None, str | None]:
     """Validate claim structure using the RuntimeClaim Pydantic model."""
     try:
         RuntimeClaim.model_validate(claim)
-        return True, None
+        return True, None, None
     except ValidationError as exc:
-        return False, str(exc)
+        return False, str(exc), _validation_error_path(exc)
 
 
 @dataclass
@@ -834,14 +850,24 @@ def verify_trace_claim(
     failure: VerificationError | None = None
     details: dict[str, str] = {}
 
-    # Step 1: Schema validation
-    schema_ok, schema_err = _validate_schema(claim_json)
-    if schema_ok:
-        verified.append("schema")
-    else:
-        unverified.append("schema")
-        failure = VerificationError.CLAIM_MALFORMED
-        details["schema_error"] = schema_err or "schema validation failed"
+    # Step 1: Schema establishment. Structural malformation wins and stops
+    # interpretation: without a valid shape the verifier has not established the
+    # bytes or fields to which a signature/key-binding verdict would refer.
+    schema_ok, schema_err, malformed_field = _validate_schema(claim_json)
+    if not schema_ok:
+        return VerificationResult(
+            status=VerificationStatus.UNVERIFIED,
+            verified_fields=[],
+            unverified_fields=["schema"],
+            failure_reason=VerificationError.CLAIM_MALFORMED,
+            attestation_age_seconds=-1,
+            is_attestation_fresh=False,
+            details={
+                "schema_error": schema_err or "schema validation failed",
+                "malformed_field": malformed_field or "claim",
+            },
+        )
+    verified.append("schema")
 
     # Step 2: Signature
     sig_ok, sig_err = _verify_signature(claim_json)
--- a/tests/unit/test_evidence_envelope_all_platforms.py
+++ b/tests/unit/test_evidence_envelope_all_platforms.py
@@ -43,7 +43,12 @@ from cmcp_runtime.audit.trace_claim import (
     generate_trace_claim,
 )
 from cmcp_runtime.tee.base import jwk_thumbprint
-from cmcp_verify.verify import ApprovedHashes, verify_trace_claim
+from cmcp_verify.verify import (
+    ApprovedHashes,
+    VerificationError,
+    VerificationStatus,
+    verify_trace_claim,
+)
 
 POLICY_HASH = "sha256:" + "a" * 64
 CATALOG_HASH = "sha256:" + "b" * 64
@@ -141,7 +146,6 @@ class _Spy:
         ("azure-cvm-sev-snp", None, "cmcp_verify.azure_cvm", "verify_azure_cvm_measurement"),
         ("sev-snp", None, "cmcp_verify.sev_snp", "verify_sev_snp_measurement"),
         ("tdx", None, "cmcp_verify.tdx", "verify_tdx_measurement"),
-        ("tdx", "opaque", "cmcp_verify.opaque", "verify_opaque_measurement"),
     ],
 )
 def test_platform_branch_reads_evidence_from_the_envelope(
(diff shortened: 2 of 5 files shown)


imran-siddique:
Batch response for this cluster is here: https://github.com/agentrust-io/agent-manifest/pull/357#issuecomment-5472317549

Short version: the finding class is real and welcome. Your CI had never run, held under first-time-contributor gating, until I released 36 runs across your PRs an hour ago, and five of your eight are now red. Please fix those, sequence trace-spec#258 against #252 which touch the same two files, and tell me the order you want them reviewed in.

altrudev:
I reconciled the released CI failure. The red step was `ruff` import formatting in the new test file, before the reproducer could execute; that is fixed at `4e2cce1`.

Because this PR is explicitly a **draft test-first reproducer**, I also changed the current-head behavior at `99ba8a3` so only the known host-language escape (`AttributeError` / `TypeError`) is reported with `pytest.xfail`. If `verify_trace_claim()` returns normally, the original `VerificationResult` and `schema` assertions run unchanged. This keeps a known-current defect from making the whole draft CI red without weakening the desired invariant or silently choosing the still-open failure-precedence rule.

I am keeping #596 out of the review queue until the implementation/precedence decision is ready. Fresh CI is currently waiting on workflow approval; no green claim yet.

imran-siddique:
Thanks for splitting this the way you did. Asserting only the uncontested invariant and holding the precedence question for a maintainer is the right call, and it is the reason this is a useful reproducer rather than a change that quietly decides something.

**The ruling: structural malformation wins.**

If the claim structure is malformed, `CLAIM_MALFORMED` is the result, and a subsequent signature or key-binding failure must not overwrite it. Encode that as first-failure-wins for the structural stage specifically, not as a general ordering over all failure kinds.

The reason is not aesthetic. A signature verdict is a statement about a specific set of bytes. When the structure did not parse, the verifier never established what those bytes were, so whatever it computed is a verdict about something it could not identify. Reporting it as `SIGNATURE_INVALID` tells the caller the signature was checked and found bad. It was not checked. That is the same defect class I have been unpicking across the other repos: absence of evidence surfacing as a negative verdict rather than as "not established". A malformed record is not a record with a bad signature; it is a record whose signature is unevaluated.

The practical consequence is the one that matters to a caller: `SIGNATURE_INVALID` says do not trust this issuer, and `CLAIM_MALFORMED` says do not trust this input. Those route to different places, and collapsing the second into the first sends operators after the wrong party.

Two things I do want from the shape of it:

1. Keep the malformed result carrying which intermediate failed to parse. Your matrix already covers `trace`, `cnf`, `jwk`, `jwk.x`, `gateway`, `audit_chain` and `attestation_evidence`; a caller debugging a producer needs to know which one, and a single flat `CLAIM_MALFORMED` throws that away.
2. Do not extend the precedence rule past the structural stage on my say-so. Ordering between signature failure and key-binding failure is a separate question with its own arguments, and this PR does not need to answer it.

Once that is in, I am happy for the xfails on the escaping shapes to become plain assertions, since the escape itself is a bug we agree on.

Separately, on CI: the one thing still red is `ruff` rule `I001` on the import block in the new test file, at head `99ba8a3`. It reports as `test (3.13, ubuntu-latest)` but the failing step is `Lint`, so nothing in your matrix has executed yet. `ruff check --fix` on that file clears it.

imran-siddique (commented):
Review/test follow-up on `141c5b2ee96c`.

The schema-first return correctly prevents malformed claims from reaching cryptographic interpretation, and the new boundary regressions pass. The full suite is not green: fresh CI reports 25 failed, 1470 passed, 6 skipped. Besides the shared SDK/TPM-root failures tracked by #616, five existing TPM tests now stop at schema validation because their claims contain `trace.runtime.raw_evidence`, which RuntimeClaim rejects. Update those positive-path fixtures to the valid claim/evidence shape and verify that their intended TPM checks still execute; preserve the early CLAIM_MALFORMED return. Local results also include a distribution metadata mismatch and an oversized-request test failure. Keep this draft until the full suite passes.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **7 failed, 1528 passed, 16 skipped, 1 warning in 64.20s (0:01:04)**.

Fresh workflow evidence: [33664858111](https://github.com/agentrust-io/cmcp/actions/runs/33664858111) (failure).

altrudev:
@imran-siddique Thank you for the review. I have replaced the intermediate compatibility workaround with the strict schema-first implementation you requested and corrected the positive-path fixtures instead. The new head is `6ffad28da75560dfba5a0c242839403f871e9c53`, reconstructed on current main `cc27f91` as one DCO-signed commit.

The original structural boundary and 15 regression tests are preserved byte-for-byte. TPM raw evidence and the separate EK chain now travel through the signed gateway envelope. The full local suite is green (1,671 passed, 14 skipped), as is the CI-scoped suite (1,629 passed, 6 skipped; 87.56% coverage), Ruff, and mypy. I also corrected the cross-platform fixture that had treated a TRACE-unsupported OPAQUE platform value as a valid positive case; supported hardware paths retain positive coverage and unsupported aliases now have explicit schema-refusal tests.

The PR body contains the exact test evidence and scope. I am keeping it draft until your review and hosted checks are complete. No manual Actions run was triggered.

imran-siddique (approved):
Reviewed 6ffad28da75560dfba5a0c242839403f871e9c53.

The schema-first return prevents malformed external structures from reaching signature/key-binding interpretation. The original 15 regressions remain, and the TPM fixtures now carry signed evidence in the gateway envelope. After installing this branch and its declared dependencies, the full local suite passed: 1,669 passed, 16 skipped. The initial failures were stale installed package metadata and an older Agent Manifest dependency. Hosted CI is also green. Leaving this PR as a draft as requested; it can move to ready when you are ready.

PR #597, opened 2026-08-30
fix(verify): classify malformed audit-bundle shapes

### LINKED ISSUE #593: verify_audit_bundle can raise on malformed bundle entry shapes instead of returning AuditBundleResult
## Current behavior

`verify_audit_bundle()` is a separately exported verifier entry point and returns `AuditBundleResult`, but at current `main` (`2c0a601805bfb31fddc49db32ae122bae8e2251e`) it assumes the externally supplied bundle and every member of `entries` already have object shape.

The first verification loop does:

```python
entries = bundle_json.get("entries", [])
...
for i, entry in enumerate(entries):
    body = {k: v for k, v in entry.items() if k != "entry_hash"}
```

So malformed but JSON-valid inputs such as:

```json
{"entries": ["unexpected"]}
```

or

```json
{"entries": [1]}
```

reach `.items()` and escape as `AttributeError` instead of producing an `AuditBundleResult(verified=False, ...)`.

A truthy non-list `entries` value also enters the loop. For example, a string is iterated character by character before failing on the first character's missing `.items()`.

The optional claim-binding half has the same structural assumption in nested reads such as:

```python
chain = claim_json.get("gateway", {}).get("audit_chain", {})
transcript = claim_json.get("trace", {}).get("tool_transcript", {})
...
x_b64 = claim_json.get("trace", {}).get("cnf", {}).get("jwk", {}).get("x", "")
```

which can raise if an intermediate value is a scalar/list rather than an object.

## Why this is distinct from #592

#592 concerns `verify_trace_claim()` continuing into structure-dependent claim verification after its Pydantic schema check has already failed.

This issue is the independent public `verify_audit_bundle()` API. It has no schema gate at all for the bundle and returns a different result type. Fixing #592 does not change this path.

## Expected invariant

Malformed external audit-bundle structure should fail closed as verifier evidence, not terminate the caller with a Python shape exception.

At minimum, `verify_audit_bundle()` should either:

1. establish the bundle/entry object and array boundaries before content verification; or
2. defensively classify malformed structure into `AuditBundleResult.failures`.

The verifier should not need to infer cryptographic meaning from a value whose JSON container type has not first been established.

## Focused regression matrix

- top-level `bundle_json` object with missing/empty `entries` -> existing `bundle has no entries` behavior preserved;
- `entries` = string/object/integer/bool -> failed `AuditBundleResult`, never an exception;
- `entries` = list containing string/integer/list/null -> failed `AuditBundleResult`, never an exception;
- one valid entry control -> existing hash/link verification unchanged;
- when `claim_json` is supplied, malformed `gateway`, `gateway.audit_chain`, `trace`, `trace.tool_transcript`, and `trace.cnf` intermediates -> failed `AuditBundleResult`, never an exception.

No wire-format or normative specification change appears necessary. This is verifier exception-safety and structural failure classification at the audit-bundle boundary.

altrudev:
## What

Malformed external audit bundles and claim containers now return a failed `AuditBundleResult` instead of escaping as Python shape exceptions. Closes #593 after review and merge.

The guards cover bundle/entry containers and nested claim objects. The follow-up also rejects non-string tool names when computing claim summaries and non-string external receipt evidence types before set membership. Hash-chain verification remains active; external receipt verification remains opt-in. Wire format and signature construction are unchanged.

## Validation

- 33 malformed-shape tests pass. Before the follow-up fix, the expanded matrix produced 10 failures and 23 passes, reproducing list/dict hashing errors and mixed-type tool-name sorting errors on correctly hashed bundles.
- 69 focused shape and external-evidence tests pass on the branch combined with current main.
- Final local unit run: 1,806 passed, 8 skipped, one package-metadata mismatch. Building and installing the matching checkout into isolated test dependencies cleared that remaining smoke test on rerun (1 passed).
- Ruff and mypy pass for the changed verifier/tests; Bandit found no issues in the changed verifier before the main merge.
- An earlier run using the older shared dependencies had 20 failures; all 20 also reproduced on the untouched PR head. The final run used isolated TRACE 0.10.0 and Agent Manifest 0.12.0 packages.

The full integration suite and hardware-dependent tests were not run locally. Hosted checks must be evaluated on the latest head, not the older green run. The PR remains draft pending independent re-review.

Current main is merged, with both changelog entries preserved. DCO sign-offs are retained. AI assistance: ChatGPT assisted with reproduction, implementation, tests, and validation; this does not establish independent security assurance.

### DIFF
--- a/src/cmcp_verify/verify.py
+++ b/src/cmcp_verify/verify.py
@@ -539,6 +539,15 @@ def _external_evidence_failure(entry_index: int, reason: str) -> str:
     return f"entry {entry_index}: {_EXTERNAL_EVIDENCE_ERROR}: {reason}"
 
 
+def _audit_bundle_shape_failure(path: str, entry_count: int = 0) -> AuditBundleResult:
+    """Classify malformed external structure without interpreting its contents."""
+    return AuditBundleResult(
+        verified=False,
+        entry_count=entry_count,
+        failures=[f"{path} has invalid object or array shape"],
+    )
+
+
 def verify_audit_bundle(
     bundle_json: dict[str, Any],
     claim_json: dict[str, Any] | None = None,
@@ -560,8 +569,39 @@ def verify_audit_bundle(
        signature). This is opt-in: receipt-less entries and callers that do not
        supply keys are unaffected, so existing evidence keeps verifying.
     """
-    failures: list[str] = []
+    if not isinstance(bundle_json, dict):
+        return _audit_bundle_shape_failure("bundle")
+
     entries = bundle_json.get("entries", [])
+    if not isinstance(entries, list):
+        return _audit_bundle_shape_failure("bundle.entries")
+    for i, entry in enumerate(entries):
+        if not isinstance(entry, dict):
+            return _audit_bundle_shape_failure(f"bundle.entries[{i}]", len(entries))
+
+    if claim_json is not None:
+        if not isinstance(claim_json, dict):
+            return _audit_bundle_shape_failure("claim", len(entries))
+        for path in (
+            ("gateway",),
+            ("gateway", "audit_chain"),
+            ("gateway", "call_summary"),
+            ("trace",),
+            ("trace", "tool_transcript"),
+            ("trace", "cnf"),
+            ("trace", "cnf", "jwk"),
+        ):
+            obj: Any = claim_json
+            for depth, name in enumerate(path):
+                if name not in obj:
+                    break
+                obj = obj[name]
+                if not isinstance(obj, dict):
+                    return _audit_bundle_shape_failure(
+                        "claim." + ".".join(path[: depth + 1]), len(entries)
+                    )
+
+    failures: list[str] = []
     if not entries:
         return AuditBundleResult(verified=False, entry_count=0, failures=["bundle has no entries"])
 
--- a/tests/unit/test_verify_audit_bundle_malformed_shapes.py
+++ b/tests/unit/test_verify_audit_bundle_malformed_shapes.py
@@ -42,7 +42,10 @@ def _one_entry_bundle() -> dict:
     ],
 )
 def test_malformed_entries_return_failed_result(entries) -> None:
-    result = verify_audit_bundle({"entries": entries})
+    try:
+        result = verify_audit_bundle({"entries": entries})
+    except (AttributeError, TypeError) as exc:
+        pytest.xfail(f"known #593 malformed-bundle escape: {type(exc).__name__}: {exc}")
 
     assert isinstance(result, AuditBundleResult)
     assert not result.verified
@@ -61,7 +64,10 @@ def test_malformed_entries_return_failed_result(entries) -> None:
     ],
 )
 def test_malformed_claim_binding_shapes_return_failed_result(claim: dict) -> None:
-    result = verify_audit_bundle(_one_entry_bundle(), claim)
+    try:
+        result = verify_audit_bundle(_one_entry_bundle(), claim)
+    except (AttributeError, TypeError) as exc:
+        pytest.xfail(f"known #593 malformed-claim escape: {type(exc).__name__}: {exc}")
 
     assert isinstance(result, AuditBundleResult)
     assert not result.verified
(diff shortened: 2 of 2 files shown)


imran-siddique:
Batch response for this cluster is here: https://github.com/agentrust-io/agent-manifest/pull/357#issuecomment-5472317549

Short version: the finding class is real and welcome. Your CI had never run, held under first-time-contributor gating, until I released 36 runs across your PRs an hour ago, and five of your eight are now red. Please fix those, sequence trace-spec#258 against #252 which touch the same two files, and tell me the order you want them reviewed in.

altrudev:
I reconciled the released CI failure here as well. Lint/type/security were already green; the red `pytest` step was the draft reproducer doing exactly what #593 reports: malformed bundle/claim containers escaped as `TypeError` / `AttributeError`.

At `e7a41a4` the tests now xfail **only when those known host-language exceptions escape**. If `verify_audit_bundle()` returns an `AuditBundleResult`, the original `verified == False` / `failures` assertions run normally. Missing-entries remains an ordinary passing control.

That keeps this draft executable without treating a known-current defect as a general CI regression, and it does not pretend #593 is implemented. I am keeping #597 out of the review queue until the narrow structural implementation is ready. Fresh CI is currently waiting on workflow approval; no green claim yet.

imran-siddique (commented):
Review/test follow-up on `e7a41a4dd7e1`.

This draft is a reproduction suite, with all 15 malformed-bundle/claim cases still xfailed on the current implementation. It therefore records the escapes but does not fix them. Fresh CI is red with 20 failed, 1439 passed, 6 skipped, 15 xfailed, including the shared SDK MISMATCH and TPM error-message failures. The local distribution-smoke failure is an existing review-environment metadata mismatch (0.4.0 runtime versus installed 0.4.1), not evidence of a new defect in these test-only changes. Keep this draft pending the #593 fix; turn the reproductions into required passing regressions when the implementation lands. The SDK compatibility follow-up is #616.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **1 failed, 1498 passed, 16 skipped, 15 xfailed, 1 warning in 73.11s (0:01:13)**.

Fresh workflow evidence: [33346156403](https://github.com/agentrust-io/cmcp/actions/runs/33346156403) (failure).

qubeena07:
Reviewed the production fix at bec8751, not the earlier reproducer commits. Found two more escape routes the shape guard at the top of verify_audit_bundle does not cover, same bug class this PR and #593 exist to close. Reproduced both directly against this branch, not just read the code.

First, around line 737:

```python
"tools_invoked": sorted(
    {
        entry["tool_name"]
        for entry in tool_calls
        if entry.get("tool_name") is not None
    }
),
```

If tool_name is a list instead of a string, this raises TypeError: unhashable type: list. Confirmed:

```python
bundle = {"entries": [{
    "entry_type": "tool_call",
    "tool_name": ["x"],
    "call_id": "c1",
    "entry_hash": "deadbeef",
    "prev_entry_hash": "genesis",
}]}
claim = {
    "gateway": {"audit_chain": {"tip": "sha256:deadbeef"}, "call_summary": {}},
    "trace": {"tool_transcript": {"hash": "sha256:deadbeef", "call_count": 1}},
}
verify_audit_bundle(bundle, claim)
# TypeError: unhashable type: 'list'
```

Second, around line 658:

```python
evidence_type = ev.get("evidence_type", "")
if evidence_type not in _EXTERNAL_EVIDENCE_TYPES:
```

Same problem, if evidence_type is a list the in check against a set raises TypeError: unhashable type: list instead of recording a failure. Confirmed the same way with an external_execution_evidence block whose evidence_type is a list.

Both reachable without a valid hash chain, since the chain check earlier in the function appends to failures and keeps going rather than returning early, so a caller does not need to get past that first to reach these two.

The top of function guard only checks that each bundle entry is a dict, not that individual fields inside a dict have the right type, so these two slip through the same way the original TypeError/AttributeError cases in #593 did before this PR's shape checks were added.

imran-siddique (approved):
Approving at `e3000fb`. Merged locally with current main, all 33 shape tests pass with no xfails, and with main's `verify.py` restored 28 of them fail, so the suite exercises the fix. Both escapes @qubeena07 found (a list-valued `tool_name` and `evidence_type`) now return `verified=False` rather than raising, and the verify and audit unit tests pass.

PR #600, opened 2026-09-01
fix(opaque): fail closed on incomplete managed appraisal

### LINKED ISSUE #594: Opaque attestation can report verified on endpoint failure or without measurement_matched
## Current behavior

At current `main` (`2c0a601805bfb31fddc49db32ae122bae8e2251e`), `cmcp_verify.opaque.verify_opaque_measurement()` starts with:

```python
result = OpaqueVerificationResult(verified=True)
```

and only flips that flag to `False` for a missing endpoint, missing evidence, or an endpoint response whose `verified` member is not `true`.

That leaves two helper-level paths where the implementation gives positive verification credit without establishing the success condition documented for Opaque Managed attestation.

### 1. Endpoint / transport / parse failure stays `verified=True`

The catch-all failure path records the error but does not clear the initial success flag:

```python
except Exception as exc:
    ...
    result.unverified_fields.append("opaque_managed_attestation")
    result.details["opaque_endpoint"] = endpoint
    result.details["opaque_error"] = type(exc).__name__

return result
```

This is not only reachable in the helper; the current unit test pins it:

```python
with patch(
    "cmcp_verify.opaque.urllib.request.urlopen",
    side_effect=OSError("timeout"),
):
    result = verify_opaque_measurement(...)

assert result.verified
assert "opaque_managed_attestation" in result.unverified_fields
```

So a timeout, TLS/network exception, malformed JSON response, or other endpoint failure produces the contradictory helper result `verified=True` while the attestation is listed as unverified.

### 2. `measurement_matched` is not part of the success predicate

`docs/spec/verification-library.md` defines the managed response as:

```text
{verified: true|false, measurement_matched: true|false, error?: string}
```

and states:

> If verified and measurement_matched: TEE identity is verified for OPAQUE Managed.

The implementation currently checks only:

```python
if body.get("verified") is True:
```

The positive test likewise returns only `{"verified": true}` and expects success. A response with `verified=true` and `measurement_matched=false`, or with the measurement result absent entirely, therefore receives the same positive helper result as one that affirmatively bound the claimed measurement.

## Integration boundary found while tracing this

`verify_trace_claim()` consumes `opaque_result.verified` as the gate for adding `hardware_attestation` to `verified_fields`, so the helper boolean is security-relevant to the enclosing verifier.

However, current claim construction also moved signed attestation bytes out of `trace.runtime` into `gateway.attestation_evidence` because TRACE `RuntimeInfo` is `extra="forbid"`. The Opaque branch still reads `_runtime.get("raw_evidence")` rather than the shared `_evidence_field(...)` path. That appears to be a separate evidence-plumbing/reachability defect: a schema-valid current claim may not reach the Opaque endpoint at all. I am not using that separate defect to strengthen this issue's exploitability claim.

This issue is therefore limited to the correctness of `verify_opaque_measurement()`'s success predicate and result state. The stale evidence source is being checked independently.

## Existing contract

This does not require a new policy decision. The original implementation issue #70 required the Opaque/TDX stubs to return partial/unverified results with descriptive `unverified_fields` when verification could not be completed rather than treating unavailable verification as success. The verification-library spec separately states the two-part Opaque success condition above.

## Expected invariant

`OpaqueVerificationResult.verified` must be true only after the configured Opaque verifier has returned an affirmative response satisfying every success predicate the profile requires. Endpoint failure, response parse failure, absent/negative measurement binding, or an explicit negative verdict must never produce positive Opaque verification credit.

## Narrow regression matrix

- endpoint returns `verified=true, measurement_matched=true` -> Opaque result verified;
- endpoint returns `verified=true, measurement_matched=false` -> not verified;
- endpoint returns `verified=true` with no `measurement_matched` -> not verified;
- endpoint returns `verified=false` -> existing failure behavior preserved;
- `urlopen` raises `OSError`/timeout -> not verified, error remains observable;
- malformed JSON response -> not verified, parse failure remains observable.

No wire-format change is required. This is a verifier success-predicate and failure-path correction.

altrudev:
Closes #594.

## What

`verify_opaque_measurement()` previously started from `verified=True`, so transport/parse exceptions could return a contradictory positive result. It also treated `{"verified": true}` as sufficient even though the managed appraisal response carries a separate `measurement_matched` predicate.

This patch follows the accepted scope from #594:

- initialize `OpaqueVerificationResult` fail-closed;
- grant positive verification only when the parsed response is an object and both `verified is True` and `measurement_matched is True`;
- reject missing, false, and non-boolean success members;
- keep endpoint-supplied failure reasons when valid;
- keep transport/parse failures unverified with an observable generic failure reason and error-type-only detail;
- refuse parsed non-object JSON explicitly rather than relying on an incidental attribute error.

## Regression coverage

The regression matrix covers:

- `verified=true` + `measurement_matched=true` -> success;
- false or missing `measurement_matched` -> refusal;
- missing `verified` -> refusal;
- `verified=false` -> refusal with endpoint failure reason preserved;
- truthy non-boolean values (`1`, `"true"`) for either predicate -> refusal;
- `null` predicates -> refusal;
- parsed non-object JSON -> refusal;
- malformed JSON -> refusal with parse error type observable;
- transport failure -> refusal with error type observable and exception text not surfaced;
- enclosing `verify_trace_claim()` regression proving a failed Opaque appraisal cannot add `hardware_attestation` to `verified_fields`.

The pre-existing tests that asserted the old fail-open behavior are corrected to the new contract.

## Scope

This intentionally does **not** change the separate stale evidence-plumbing path (`trace.runtime.raw_evidence` versus the shared evidence helper). No wire-format or policy decision is introduced here.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial-case design, implementation drafting, and diff review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/src/cmcp_verify/opaque.py
+++ b/src/cmcp_verify/opaque.py
@@ -50,11 +50,12 @@ def verify_opaque_measurement(
     header. The header value is never logged -- _redact_auth_headers() strips it
     before any debug output (HW-008).
     """
-    result = OpaqueVerificationResult(verified=True)
+    # Fail closed by default. Positive verification credit is granted only after
+    # the managed verifier explicitly confirms both required success predicates.
+    result = OpaqueVerificationResult(verified=False)
 
     endpoint = opaque_endpoint or os.environ.get(_OPAQUE_ENDPOINT_ENV)
     if not endpoint:
-        result.verified = False
         result.failure_reason = "opaque_endpoint_not_configured"
         result.unverified_fields.append("opaque_managed_attestation")
         result.details["hint"] = (
@@ -64,7 +65,6 @@ def verify_opaque_measurement(
 
     if raw_evidence is None:
         # Fail closed: an attestation claim with no evidence cannot verify.
-        result.verified = False
         result.failure_reason = "no_raw_evidence"
         result.unverified_fields.append("opaque_managed_attestation")
         result.details["opaque_endpoint"] = endpoint
@@ -101,12 +101,21 @@ def verify_opaque_measurement(
         with urllib.request.urlopen(req, timeout=_OPAQUE_TIMEOUT_SECONDS) as resp:  # nosec B310 - req is a Request object with explicit HTTPS endpoint
             body = json.loads(resp.read().decode())
 
-        if body.get("verified") is True:
+        if not isinstance(body, dict):
+            result.failure_reason = "opaque_invalid_response"
+            result.unverified_fields.append("opaque_managed_attestation")
+            result.details["opaque_response_type"] = type(body).__name__
+        elif body.get("verified") is True and body.get("measurement_matched") is True:
+            result.verified = True
             result.verified_fields.append("opaque_managed_attestation")
             result.details["opaque_endpoint"] = endpoint
         else:
-            result.verified = False
-            result.failure_reason = body.get("failure_reason", "opaque_verification_failed")
+            failure_reason = body.get("failure_reason")
+            result.failure_reason = (
+                failure_reason
+                if isinstance(failure_reason, str) and failure_reason
+                else "opaque_verification_failed"
+            )
             result.unverified_fields.append("opaque_managed_attestation")
             result.details["opaque_response"] = str(body.get("details", ""))
 
@@ -119,6 +128,7 @@ def verify_opaque_measurement(
             type(exc).__name__,
             _redact_auth_headers(request_headers),
         )
+        result.failure_reason = "opaque_verification_error"
         result.unverified_fields.append("opaque_managed_attestation")
         result.details["opaque_endpoint"] = endpoint
         result.details["opaque_error"] = type(exc).__name__
--- a/tests/unit/test_opaque_fail_closed_594.py
+++ b/tests/unit/test_opaque_fail_closed_594.py
@@ -1,0 +1,143 @@
+"""Regression matrix for Opaque managed-attestation fail-closed semantics (#594)."""
+from __future__ import annotations
+
+import json
+from unittest.mock import MagicMock, patch
+
+import pytest
+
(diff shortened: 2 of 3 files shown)


altrudev (commented):
@carloshvp Implemented the accepted #594 boundary in this PR against current main. The exact head is `74df873d0394518953ac78fa2824fca1a1f5e5d2`; the patch keeps the stale evidence-plumbing path out of scope and adds the full fail-closed matrix plus the enclosing `hardware_attestation` regression. Upstream workflows are currently `action_required` pending repository approval rather than reporting test failures.

imran-siddique (approved):
Merging. Two defects, and the first is the one that matters.

**Starting from `verified=True` inverts the direction the whole function should fail in.** A transport or parse exception then returns a contradictory positive: the code never reached a verification, and the result says it verified. Initializing `OpaqueVerificationResult` fail-closed makes the positive something that has to be earned rather than something that has to be revoked, which is the difference between a check and a formality.

**The second is subtler and I am glad you did not stop at the first.** Treating `{"verified": true}` as sufficient while the managed appraisal response carries a separate `measurement_matched` predicate means a response could report the appraisal succeeded without the measurement having matched. Requiring both, and rejecting missing, `false` and non-boolean members rather than coercing them, is the right reading: two predicates in a response are two questions, and answering one is not answering both.

Refusing parsed non-object JSON explicitly, rather than relying on an incidental attribute error, is the same discipline that has been running through this codebase all week: a refusal you wrote is a decision, and a refusal that falls out of an `AttributeError` is an accident that happens to work today.

Keeping endpoint-supplied failure reasons when valid, while holding transport failures to a generic reason with error-type-only detail, is the correct split. An endpoint's own explanation is useful; an exception string from our side of the wire is an information leak dressed as diagnostics.

Eleven checks, only the maintainer gate red. Closes #594.

PR #601, opened 2026-09-02
refactor(tpm): delegate NV certify parsing to Agent Manifest

noah-ing:
## What

- Retire cMCP's duplicate `TPMS_ATTEST` / `TPMS_NV_CERTIFY_INFO` wire parser and delegate parsing to `agent_manifest.parse_tpm_nv_certify`.
- Preserve cMCP's public `parse_nv_certify()` return shape and `ValueError` contract while using the canonical parser's exact-buffer and union-type checks.
- Verify AK signatures over the canonical inner `TPMS_ATTEST` bytes when evidence arrives in `TPM2B_ATTEST` transport framing.
- Add a checksummed two-certify `swtpm` reference pair that exercises the released parser, certificate-chain check, TPM signature envelopes, phase bindings, index identity, and extend relation end to end.

## Why

Agent Manifest PR agentrust-io/agent-manifest#304 added the canonical NV-certify parser in release 0.11.2 and explicitly enabled cMCP to retire its remaining local wire parser. cMCP already requires `agent-manifest>=0.11.2`, so keeping both implementations only preserves a trust-boundary split.

The implementations had also developed observable framing differences: the local parser rejected valid size-prefixed `TPM2B_ATTEST` input and did not require `nvContents` to consume the complete signed structure. This completes the downstream migration without introducing a new dependency or public API.

## Security impact

TPM wire parsing now has one authority across Agent Manifest and cMCP. For size-prefixed transport, signature verification covers `parsed.attest.raw`—the inner structure signed by the AK—not the unsigned two-byte transport length. Quote/NV type confusion, truncation, and undeclared trailing data remain fail-closed.

cMCP still owns the gateway-specific two-certify phase binding, same-index check, expected gateway digest, and extend relation. The configured AK trust-root and signature requirements are unchanged.

The reference corpus is software-TPM evidence under a synthetic test CA. It does not establish physical TPM or TEE provenance, vendor enrollment, EK-to-AK activation, hardware key residency, requester/TPM co-location, boot state, or runtime integrity.

## Test plan

- [x] `pytest` passes — 1,474 passed, 6 pre-existing environment-gated hardware skips; 86.95% coverage in an `autocrlf=true` clean checkout.
- [x] `ruff check` passes.
- [x] `mypy` passes across 65 source files.
- [x] Manual test performed (described below).

Additional verification:

- 33 focused tests pass on Python 3.11, 3.12, and 3.13.
- Bandit reports no issues; `pip-audit --skip-editable` reports no known dependency vulnerabilities.
- Wheel and sdist build, pass `twine check`, install independently, import outside the checkout, and pass the installed CLI smoke.
- Strict MkDocs build passes.
- No-cache Docker build passes; the runtime imports Agent Manifest 0.11.2 and runs as UID/GID 10001.
- Gitleaks reports zero findings; detect-secrets passes after excluding the documented public TPM fixture bytes and qualifying-data digests.

Manual producer path:

1. Started a fresh `swtpm` 0.7.3 instance and created a restricted RSA-2048 RSASSA/SHA-256 AK with `tpm2-tools` 5.6.
2. Defined and seeded a SHA-256 `TPM_NT_EXTEND` index, certified its pre-value, extended the gateway digest, and certified its post-value.
3. Wrapped the exact AK public key in a synthetic test certificate, then removed all private CA material, TPM state, and object contexts.
4. Verified both bare and size-prefixed forms, exact index Name and contents, `post == SHA256(pre || gateway_digest)`, replay rejection under a new nonce, signature-tamper rejection, and every committed artifact checksum.

## DCO sign-off


imran-siddique (approved):
Approved. The migration to the Agent Manifest parser is a clean delegation, the swtpm reference corpus and the expanded unit coverage are welcome additions, and all functional checks are green. Thanks Noah.

noah-ing:
Thank you for the detailed review and for carrying this through #602 and the 0.4.1 release. I especially appreciated the careful scope correction and the authorship credit. Working through the verifier boundary and reproducing the issue concretely taught me a great deal.

PR #602, opened 2026-09-02
fix(tpm): bind NV appraisal to trusted policy

imran-siddique:
Fixes the high-severity issue tracked as GHSA-943q-hvhp-mrx2, reported and fixed by @noah-ing.

## Problem

`verify_gateway_measurement()` could return `verified=True` for a correctly signed TPM2_NV_Certify pair that does not refer to cMCP's configured `TPM_NT_EXTEND` object or its complete 32-byte range. The verifier compared the two signed TPM Names only with each other, never against trusted verifier policy, and treated an evidence-carried gateway digest as sufficient when no expected digest was configured. No signature forgery is needed to exploit it.

## Fix

- Adds a mandatory verifier-owned `GatewayNvAppraisalPolicy`: exact authorized written TPM Name, offset zero, 32-byte extent, expected gateway digest.
- Derives the production Name by dependency-free marshaling of cMCP's `TPMS_NV_PUBLIC` template.
- Requires both signed Names, offsets, extents, the envelope digest, transcript bindings, signatures, and the extend relation to match policy.
- Makes digest-only legacy calls fail closed.
- Validates the complete producer-side public area and TPM-returned Name before any read, extend, or certify, across the unwritten and written state transitions.
- Maps only an actual `TPM_RC_NV_UNINITIALIZED` read to the initial zero value.

## Breaking change

Callers that omitted the full appraisal policy now fail closed. This is intentional.

## Scope limit

The startup NV-certify pair is not carried by the ordinary TRACE schema and `verify_trace_claim()` does not call `verify_gateway_measurement()`, so this was a false positive in the standalone verifier, not an end-to-end bypass in the ordinary request path. That is why it is scored high rather than critical.

## Verification

Genuine swtpm corpus covering the canonical production object, a correctly signed same-handle ordinary object with an arranged byte relation, and a correctly signed partial range. Live Ubuntu 24.04 swtpm plus tpm2-pytss run through `extend_gateway_measurement()`: written attributes `0x22060042`, returned Name matched the expected value, and `TPM_RC_NV_UNINITIALIZED` mapped to zero.

Rebased from the advisory's private fork onto main after #601 merged. Commit authorship is Noah's.


imran-siddique:
Windows CI caught a packaging gap in the fixture corpus, not a defect in the fix.

`.gitattributes` covers `swtpm-nv-certify-0.7.3` from #601, but the new `swtpm-nv-policy-0.7.3` corpus was never added. Git therefore checked its text vectors out with CRLF on Windows and `test_swtpm_nv_policy_artifact_checksums` failed against its own SHA256SUMS. The seven affected files were `ak-cert.pem`, `ak-public.pem`, `synthetic-root.pem`, `producer-versions.txt`, and both `nv-public.yaml` copies; the `.bin` vectors survived because Git auto-detects them as binary.

Amended the same `-text` / `binary` / `whitespace=-blank-at-eof` rules onto the policy corpus, adjusted for its per-case subdirectories. Verified on Windows: all 25 policy vectors and the certify corpus now hash byte-exact. Amended rather than added as a second commit so the squash merge keeps Noah as author.

PR #603, opened 2026-09-02
test: preserve NV fixture bytes on Windows

noah-ing:
## What

Preserve the new `swtpm-nv-policy-0.7.3` reference corpus byte-for-byte on every checkout by marking the directory `-text`, matching the existing `swtpm-nv-certify-0.7.3` corpus policy.

## Why

PR #602's Windows jobs check out six checksummed text artifacts with CRLF endings. The first mismatch is `ak-cert.pem`, so the checksum regression stops there, but the PEM, YAML, and producer-version files are all evidence bytes and must not be rewritten by checkout configuration.

This changes no fixture, expected digest, production code, or security behavior. It fixes checkout normalization rather than accepting platform-specific hashes.

## Verification

- Fresh clone with `core.autocrlf=true`: all six affected artifacts report `i/lf w/lf attr/-text` under `git ls-files --eol`.
- Both genuine signed-corpus integration suites pass: **10 passed**.
- Every entry in the new corpus `SHA256SUMS` verifies.
- `git diff --check` passes.

This PR intentionally targets `security/nv-appraisal-policy`; merging it advances and reruns #602 without creating an unrelated change on `main`.

## DCO

- [x] Signed off under the Developer Certificate of Origin.


noah-ing:
Superseded by upstream commit 3f63315f31eee21e29e8c6231802fc8907e38fee on #602, which adds the same recursive `-text` guarantee plus explicit nested binary and YAML whitespace attributes. Closing this stacked duplicate to keep review focused on #602.

PR #606, opened 2026-09-03
feat(execution): non-operational execution-state foundation (#565)

### LINKED ISSUE #565: Define session-independent execution correlation for post-session MCP

Yatsuiii:
This PR retains standalone execution-registry primitives and a typed audit `execution_id` field. Execution correlation is unavailable in the running gateway.

Valid supplied execution IDs receive an audited refusal before health checks, catalog lookup, or upstream discovery. Malformed IDs, including explicit null, are rejected. Calls omitting the field retain the existing execution path.

Production registry wiring, admission/finalization calls, and provisional action-binding code are removed. Opaque binding storage/comparison remains only in the disconnected registry; it does not establish action equivalence.

Activation requires both #588’s adopted action-binding contract and requirement 7’s atomic terminal-state/audit persistence boundary, with integrated crash/recovery tests. This PR does not close #565.

Validation: focused tests, Ruff and MyPy passed. The full suite has one startup-test failure also reproduced on clean main.


Yatsuiii:
Thanks for catching this. I missed that the HTTP handler treated a present non-string execution_id the same as an omitted one, which skipped execution-ID validation. The added HTTP tests cover the gap my proxy tests missed.

imran-siddique (commented):
Review/test follow-up on `2579f01c6ecf`.

The malformed execution-ID ingress repair remains covered, and the full local suite passed with the existing review environment. Fresh CI still fails: 20 failed, 1634 passed, 6 skipped on Python 3.13/Linux, principally Agent Manifest SDK verification returning MISMATCH, plus the TPM root-error assertion. Other matrix jobs were cancelled by fail-fast. Follow the SDK compatibility work in #616 and rerun with fresh dependencies after it lands. The provisional action binding (#588) and separate audit/registry transactions are explicitly incomplete requirements; keep that scope visible before requesting final approval. This is not ready to merge.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **1625 passed, 16 skipped, 1 warning in 88.82s (0:01:28)**.

Fresh workflow evidence: [34012436620](https://github.com/agentrust-io/cmcp/actions/runs/34012436620) (failure).

Yatsuiii:
Merged main through #617 into this branch. That brings in #616 and CI is green on all six matrix rows. The Agent Manifest MISMATCH failures and the TPM root-error assertion are both gone.

The merge doesn't touch any of the 11 files here. Diff against main is still +1271/-3, and `2579f01` is unchanged.

Still incomplete, same as the PR description:

- Action binding is a stub. `admit()` takes an opaque digest and never computes it. `provisional_action_binding` reuses the catalog canonicalizer without deciding the preimage, that's #588.
- The execution row and the terminal audit entry are in separate SQLite databases, so requirement 7's single transaction boundary isn't met. The proxy writes the audit entry first, then finalizes. What's left is an audit entry showing a definite outcome while the registry says `outcome_unknown`.

Ready for review with that scope.

imran-siddique (changes requested):
Reviewed 77b805746b30740c0b2d3cd9cc13c8cd7fd32c80.

The current head has passing hosted CI across all six matrix rows, so the earlier dependency-related test failures are cleared. The two implementation gaps you identified still block merging this as the execution-registry implementation: admission accepts an opaque action digest without establishing its preimage, and terminal audit publication and registry finalization are separate transactions. A crash between them can leave a definite audit outcome alongside outcome_unknown. Please complete the agreed action-binding contract (#588) and requirement 7, or narrow this PR to an explicitly non-operational foundation with the incomplete execution path unavailable. Passing CI does not resolve those contract gaps.

Yatsuiii:
Took the non-operational foundation option. The incomplete execution path is unavailable, and the provisional action-binding implementation and production registry wiring are removed. The standalone registry still stores and compares opaque bindings, but no gateway call uses it. Both #588 and requirement 7 remain prerequisites for activation.

qubeena07 (approved):
I read through the full diff along with the earlier review thread before writing this up. The move to a non operational foundation addresses both change requests from before. I checked the code directly and confirmed that nothing in the production path, not proxy.py, not server.py, not startup, calls admit, finalize, or recover, or constructs an ExecutionRegistry at all. Only the test suite does, and there's now a dedicated test guarding against that changing by accident.

The refusal logic in the proxy is ordered correctly, running before catalog lookup and before any upstream discovery. It also draws the right distinction: a malformed or null execution id gets refused and is never written into the audit record, while a valid but unsupported one gets refused and is recorded. Both paths have direct test coverage, including the HTTP boundary case.

I don't see anything blocking here. Approving with the expectation that the action binding contract and the atomic audit and registry boundary stay required before this gets wired into anything live.

imran-siddique (approved):
Approving. You took the second option from my 7 September review and the narrowing is real, so this lands as a foundation rather than as the execution-registry implementation.

I checked the code rather than the section that describes it, because "non-operational" is a claim about reachability and the only way to see it is to look for the callers.

`ExecutionRegistry` now appears nowhere in `src/cmcp_runtime/` outside its own package export. No construction, no `admit()`, no `finalize()`, no `recover()` in the gateway. `binding.py` is deleted and the CLI wiring with it, so there is no runtime opt-in to turn on by configuration either. That is what makes the two contract gaps I named harmless here: an admission path that accepts an opaque digest without establishing its preimage cannot be reached, and a crash between terminal audit publication and registry finalization cannot occur when nothing finalizes.

The ingress refusal is at the right place. `_check_execution_available` runs as **Step 1** in the call path, before the health check, before catalog drift, and before any upstream discovery, with `entry=None` because no catalog entry has been consulted yet. Three behaviours, each as documented: a valid supplied `execution_id` is refused `execution_correlation_unavailable` with `finalization.execution_id` set first, so the audit retains the asserted identifier; a malformed one is refused `execution_invalid_execution_id` without that assignment, so a bad value never enters audit identity; omission returns `None` immediately and the legacy path is untouched. Step 0 serializes and hashes the request before any early refusal, so a refusal still carries a stable request hash with a null server.

CI is green on all twelve real checks across six matrix rows. The red `gate` in the rollup is a stale entry from before the approvals, with a later `gate=SUCCESS` beside it.

The remaining work is unchanged and now has a clean boundary: #588 defines the canonical action binding, and a single-writer store is needed before terminal audit and registry finalization can share a transaction. Neither is owed by this PR any more.

PR #608, opened 2026-09-05
fix(verify): bind platform evidence to the TRACE nonce

### LINKED ISSUE #595: Non-TPM hardware verifier branches still read evidence from forbidden trace.runtime fields after #469
## Regression boundary

#469 moved signed platform evidence to `gateway.attestation_evidence` because canonical TRACE `RuntimeInfo` is `extra="forbid"`. That was the correct boundary and it made the TPM verifier reachable from a schema-valid current claim.

At current `main` (`2c0a601805bfb31fddc49db32ae122bae8e2251e`), the producer applies that relocation to every `AttestationReportInfo`:

```python
class AttestationEvidence(BaseModel):
    raw_evidence: str | None = None
    quote_signature: str | None = None
    cert_chain: str | None = None
    ek_cert_chain: str | None = None
...
gateway = GatewayAddenda(
    ...,
    attestation_evidence=_build_evidence(attestation_report),
)
```

and `_build_runtime()` now emits only the canonical TRACE runtime fields.

TRACE `RuntimeInfo` permits `platform`, `measurement`, `rim_uri`, `nonce`, and `firmware_version`; it does not permit `raw_evidence`, `cert_chain`, or quote material.

The verifier has a shared `_evidence_field()` helper for exactly this transition:

```python
evidence = claim_json.get("gateway", {}).get("attestation_evidence") or {}
return _decode_evidence(evidence.get(name) or runtime.get(name))
```

but only the TPM branch consistently uses it. Three shipped/current hardware branches still read the old location.

## Azure CVM: schema-valid current claim can reach `KeyError`

The current Azure provider returns its HCL/SNP + AK-quote envelope as `AttestationReport.raw_evidence`. Claim generation therefore places it at `gateway.attestation_evidence.raw_evidence`.

The verifier instead does:

```python
raw_bytes = base64.b64decode(_runtime["raw_evidence"])
```

A claim generated by current code has no such runtime member, so this is not merely absent evidence: the branch indexes a key that canonical `RuntimeInfo` cannot contain.

## SEV-SNP: the raw report is also stranded, not only the known cert-chain gap

The SEV-SNP collector returns the genuine SNP report as `AttestationReport.raw_evidence`. Current verification reads:

```python
raw_ev = _runtime.get("raw_evidence")
raw_bytes = base64.b64decode(raw_ev) if raw_ev else None
...
_chain_b64 = _runtime.get("cert_chain")
```

so a schema-valid current claim passes `None` as the report bytes even when `gateway.attestation_evidence.raw_evidence` contains them.

`LIMITATIONS.md` already discloses that the SEV-SNP `cert_chain` path was not migrated in #469. I am **not** reporting that existing disclosure as a new finding. The additional problem here is that the raw SNP report itself follows the same stale read, which prevents the lower-level verifier from appraising the evidence that the current producer actually emitted.

## Intel TDX: current TDREPORT is likewise read from the stale location

`TDXProvider.get_attestation_report()` returns the 1024-byte TDREPORT as `raw_evidence`. Claim generation moves it to `gateway.attestation_evidence.raw_evidence`, while verification does:

```python
raw_ev = _runtime.get("raw_evidence")
raw_bytes = base64.b64decode(raw_ev) if raw_ev else None
```

so the current live TDREPORT is not passed to `verify_tdx_measurement()`.

The later `raw_quote` path is a separate capability question: `AttestationReportInfo` / `AttestationEvidence` currently have no `raw_quote` member and the live `TDXProvider` collects a TDREPORT, not a DCAP quote. This issue does not propose inventing a new quote carrier; it is limited to making evidence that the current producer already emits reach the verifier.

## Why this is distinct from #370

#370 was the report-signature hardening task. #469's commit message explicitly says evidence had to move out of `trace.runtime`, fixes TPM to consume the new envelope, and notes that the SEV-SNP cert-chain migration was not included.

This issue is the integration invariant after that move:

> If current claim generation emits platform evidence into `gateway.attestation_evidence`, every verifier branch consuming that existing evidence must read the same committed location (with legacy fallback only where intentionally supported).

Otherwise producer and verifier agree on the field semantically but cannot exchange it structurally.

## Narrow fix direction

Reuse `_evidence_field()` rather than adding provider-specific accessors:

- Azure CVM: obtain `raw_evidence` through `_evidence_field()`; never direct-index a forbidden runtime member.
- SEV-SNP: obtain current `raw_evidence` through `_evidence_field()`; the already-known `cert_chain` limitation can be migrated in the same direction if maintainers want that included.
- TDX: obtain the current TDREPORT `raw_evidence` through `_evidence_field()`; do not broaden this change into defining a new DCAP `raw_quote` producer contract.

## Regression evidence to pin

For each supported branch, build the claim through `generate_trace_claim()` rather than hand-inserting legacy runtime fields, then assert:

- `trace.runtime` contains no non-TRACE evidence members;
- `gateway.attestation_evidence.raw_evidence` contains the emitted evidence;
- Azure verification does not raise `KeyError` and the lower verifier receives the decoded envelope;
- SEV-SNP lower verification receives the decoded SNP report rather than `None`;
- TDX lower verification receives the decoded TDREPORT rather than `None`;
- legacy fallback behavior, if retained, remains explicit rather than becoming the only working path.

No TRACE schema or wire-format expansion is required for the existing `raw_evidence` fix. The carrier already exists in cMCP; this is producer/verifier plumbing consistency.

carloshvp:
## What

Bind SNP, Azure CVM and TDX evidence to the 64-byte value already serialized as `trace.runtime.nonce`. These branches previously read `trace.runtime.report_data`, which the canonical schema forbids, and passed `None` to lower-level APIs where it disables the binding check. The verifier now decodes the canonical nonce, supplies its hex representation to platform appraisal, and refuses hardware credit when the expectation is absent or malformed.

## Why

Addresses the report-data portion of #595 after #607 restored evidence routing. No new report-data carrier is needed: `_build_runtime()` already encodes `AttestationReportInfo.report_data` into `trace.runtime.nonce`.

For example, a schema-valid, correctly signed SNP claim can currently contain a genuine synthetic VCEK-signed report committing a different nonce and still receive `hardware_attestation` credit. With this change, the same claim returns `HARDWARE_ATTESTATION_FAILED` and no hardware credit. Azure's comparison checks the AK-signed quote's `extraData == SHA-256(nonce)`; TDX checks TDREPORT report-data and retains partial status without an authenticated quote.

This PR deliberately does not close #595. TDX DCAP quote collection and claim transport remain absent; the current provider collects only a TDREPORT. `LIMITATIONS.md` records that remaining work and removes the obsolete statement that SNP certificate-chain routing is still missing.

## Security impact

Restores the link between the signed hardware evidence and the claim's gateway-key/audit-root nonce. Generic key/root consistency checks within a claim do not establish this link by themselves. Missing, short, oversized, illegal-alphabet and noncanonical nonce encodings cannot turn the platform's binding check off. Existing failure precedence and partial-verification reporting are preserved.

Tests use locally generated synthetic SNP/Azure certificate chains and signatures, plus synthetic TDX TDREPORTs. No new live-hardware validation is claimed. The standalone platform verifier APIs are unchanged.

## Test plan

- [x] Unit, conformance and integration suite: **1577 passed, 6 skipped**, **87.72% coverage**.
- [x] All **30 new regressions fail on base `7178c4089a022d93451077aff7ec090759c45b7a`** and pass with this change. They cover matching reports, mismatches in both nonce halves, and invalid expectations across SNP, Azure and TDX.
- [x] Existing five evidence-envelope routing regressions pass.
- [x] `ruff check src/ tests/`
- [x] `mypy src/cmcp_runtime/ src/cmcp_verify/` (66 source files)
- [x] `bandit -r src/ -c pyproject.toml`
- [x] `pip-audit --skip-editable` (no known dependency vulnerabilities)
- [x] `git diff --check`

## DCO sign-off

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- Bind SNP, Azure CVM and TDX evidence to the claim's existing
+  `trace.runtime.nonce` (#595). These branches previously read a `report_data`
+  field forbidden by the runtime schema, passing `None` to an optional binding
+  check. A valid report for a different key or audit root could therefore receive
+  hardware-verification credit on the SNP and Azure paths. The verifier now
+  requires the canonical 64-byte nonce and checks it against the report (or the
+  AK-signed quote on Azure). TDX TDREPORT-only claims remain partially verified;
+  this does not add DCAP quote collection or transport.
+
 ## [0.4.1] - 2026-09-02
 
 **Anyone running 0.4.0 should upgrade.** On 0.4.0 `verify_gateway_measurement()` could return
--- a/LIMITATIONS.md
+++ b/LIMITATIONS.md
@@ -106,9 +106,14 @@ Three gaps are worth stating plainly for the TPM path:
   rejected as `CLAIM_MALFORMED` before platform verification runs. Signed
   evidence therefore travels as `gateway.attestation_evidence`, a cmcp-owned
   field. The verifier still reads `trace.runtime` as a fallback so older claims
-  keep working, but that path cannot pass schema validation. **The SEV-SNP
-  `cert_chain` path has not been migrated and remains subject to this**, so its
-  VCEK chain verification does not engage for a schema-valid claim.
+  keep working, but that path cannot pass schema validation. All current
+  platform branches read the envelope, including the SNP VCEK chain. SNP,
+  Azure CVM and TDX compare their evidence binding against the 64-byte value
+  already carried as `trace.runtime.nonce`; missing or malformed nonces cannot
+  disable that check. **TDX DCAP quote collection and transport remain absent**:
+  the current provider collects a TDREPORT, and the claim models do not carry
+  `raw_quote`. TDX claims therefore remain partially verified without a verified
+  quote signature. The standalone quote-verification API is a separate path.
 - **The gateway NV-certify pair is not an ordinary TRACE-claim property.** TPM
   startup collects a bracketing pair, and the standalone appraisal accepts it
   only with a verifier-owned root, nonce, exact NV Name and range, and expected
--- a/docs/spec/attestation.md
+++ b/docs/spec/attestation.md
@@ -284,6 +284,14 @@ nonce = JWK_thumbprint(tee_public_key) (32 bytes) || random_salt (32 bytes)
 - `random_salt`: 32 random bytes generated once per enclave startup, so two enclave instances produce distinct nonces even with the same key (e.g. blue-green deploy). On SEV-SNP, TDX and Azure CVM this half carries the gateway measurement instead; see §3.3.2.
 - The 64-byte value is passed as the `report_data` / `user_data` / `reportdata` / `qualifying_data` field when requesting the hardware attestation report. The field name varies by provider; the semantic is the same: a caller-supplied value included in the signed measurement.
 
+In a canonical TRACE Claim this value is serialized as `trace.runtime.nonce`
+(unpadded base64url), not as a separate `trace.runtime.report_data` field.
+SNP, Azure CVM and TDX verification decode that exact 64-byte nonce and compare
+it with the platform evidence. An absent or malformed nonce fails hardware
+appraisal rather than skipping the binding check. On Azure the comparison is
+against the AK-signed quote's `extraData`, which commits `SHA-256(nonce)`;
+the paravisor controls the SNP report's own `REPORT_DATA`.
+
 Verifier check (key binding, CRYPTO-001):
 
 ```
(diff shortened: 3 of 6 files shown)


imran-siddique (approved):
Reviewed at f2b5e55ec7e55016288997dff312e897fa781cd0. The 30 new regressions fail against the base verifier and pass with this change. The current head also has a successful Linux/Windows, Python 3.11–3.13 CI matrix.

The dispatch now requires the full nonce expectation before platform appraisal, so an absent or malformed value cannot disable the binding comparison. The SNP/Azure mismatch controls and TDX partial-verification behavior are covered.

The canonical unpadded nonce requirement is an intentional cMCP representation restriction. Current producers emit that form; equivalent padded or noncanonical encodings are rejected. The hardware comparison itself remains a comparison of decoded bytes. TDX quote collection and transport remain outside this change.

PR #627, opened 2026-09-10
docs(spec): define execution action binding v1

### LINKED ISSUE #588: Define canonical bytes for execution action/intent bindings

altrudev:
Closes #588.

## What

Records the five points Imran ruled on in #588 as one normative execution-action binding contract:

1. RFC 8785/JCS canonicalization over UTF-8 bytes.
2. Algorithm-prefixed lowercase-hex digest representation; `sha256` and `sha384` accepted, other algorithms refused.
3. Action preimage fields `agent_id`, `action_type`, `action_scope`, and `action_timestamp`.
4. Domain separation inside the hashed bytes using `domain: "cmcp.execution-action-binding"`.
5. Version discrimination inside the hashed bytes using `version: 1`.

The document keeps correlation, logical-operation identity, replay permission, and external outcome evidence distinct. Same binding may enter retry/correlation classification subject to terminal-state rules; a changed binding under the same `(authenticated agent identity, execution_id)` is refused before upstream invocation.

## Vectors

Adds a companion machine-readable vector set covering:

- ASCII JCS + SHA-256;
- non-ASCII UTF-8 (`工具`) + SHA-256;
- domain separation;
- version separation;
- SHA-384 rendering;
- unsupported algorithm refusal;
- the existing `1` versus `1.0` semantic-adjudication witness, explicitly retained as a governing-semantics question rather than automatic proof the binding must distinguish the two.

The expected JCS bytes and digests are fixed in the file so an independent implementation can reproduce them without using the implementation under test as its oracle.

## Scope

Docs/spec and vectors only. No runtime activation, registry integration, admission-path change, terminal-state/audit transaction change, or exactly-once claim. #565 remains the owner of integration, durable terminal/audit consistency, crash/recovery, and replay evidence.

Base: `69355037325927f28045686d548d3ea78d80ab6e` (the exact `main` revision Imran used for the ruling).

DCO: signed off. AI-assistance disclosure: ChatGPT assisted with ruling reconciliation, contract drafting, vector construction, and exact-revision preparation. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/docs/spec/execution-action-binding-v1-vectors.json
+++ b/docs/spec/execution-action-binding-v1-vectors.json
@@ -1,0 +1,104 @@
+{
+  "profile": "cmcp.execution-action-binding-vectors/1",
+  "contract": "docs/spec/execution-action-binding.md",
+  "vectors": [
+    {
+      "id": "EAB-001-ascii-sha256",
+      "preimage": {
+        "domain": "cmcp.execution-action-binding",
+        "version": 1,
+        "agent_id": "spiffe://agent.example/payments/prod",
+        "action_type": "transfer",
+        "action_scope": "account:merchant-7",
+        "action_timestamp": "2026-09-10T16:30:00Z"
+      },
+      "jcs_utf8": "{\"action_scope\":\"account:merchant-7\",\"action_timestamp\":\"2026-09-10T16:30:00Z\",\"action_type\":\"transfer\",\"agent_id\":\"spiffe://agent.example/payments/prod\",\"domain\":\"cmcp.execution-action-binding\",\"version\":1}",
+      "algorithm": "sha256",
+      "binding": "sha256:305b05b249b159fa9dea8cbbfaa0c19b71aa867a6da718693c1c7d4b83f52fe4",
+      "expected": "accept"
+    },
+    {
+      "id": "EAB-002-nonascii-sha256",
+      "preimage": {
+        "domain": "cmcp.execution-action-binding",
+        "version": 1,
+        "agent_id": "spiffe://agent.example/payments/prod",
+        "action_type": "transfer",
+        "action_scope": "工具/merchant-7",
+        "action_timestamp": "2026-09-10T16:30:00Z"
+      },
+      "jcs_utf8": "{\"action_scope\":\"工具/merchant-7\",\"action_timestamp\":\"2026-09-10T16:30:00Z\",\"action_type\":\"transfer\",\"agent_id\":\"spiffe://agent.example/payments/prod\",\"domain\":\"cmcp.execution-action-binding\",\"version\":1}",
+      "algorithm": "sha256",
+      "binding": "sha256:6c6a43dd37e7762b65063a24fcbf77bb914f95e1ab976405e3e2cafa54b1d08a",
+      "expected": "accept"
+    },
+    {
+      "id": "EAB-003-domain-separation",
+      "preimage": {
+        "domain": "cmcp.other-binding",
+        "version": 1,
+        "agent_id": "spiffe://agent.example/payments/prod",
+        "action_type": "transfer",
+        "action_scope": "account:merchant-7",
+        "action_timestamp": "2026-09-10T16:30:00Z"
+      },
+      "jcs_utf8": "{\"action_scope\":\"account:merchant-7\",\"action_timestamp\":\"2026-09-10T16:30:00Z\",\"action_type\":\"transfer\",\"agent_id\":\"spiffe://agent.example/payments/prod\",\"domain\":\"cmcp.other-binding\",\"version\":1}",
+      "algorithm": "sha256",
+      "binding": "sha256:c9fc3b387ec2cf3fc981210a81d6dee3544e272a49c53f836abc726bbfac655a",
+      "expected": "reject_as_v1_execution_action_binding"
+    },
+    {
+      "id": "EAB-004-version-separation",
+      "preimage": {
+        "domain": "cmcp.execution-action-binding",
+        "version": 2,
+        "agent_id": "spiffe://agent.example/payments/prod",
+        "action_type": "transfer",
+        "action_scope": "account:merchant-7",
+        "action_timestamp": "2026-09-10T16:30:00Z"
+      },
+      "jcs_utf8": "{\"action_scope\":\"account:merchant-7\",\"action_timestamp\":\"2026-09-10T16:30:00Z\",\"action_type\":\"transfer\",\"agent_id\":\"spiffe://agent.example/payments/prod\",\"domain\":\"cmcp.execution-action-binding\",\"version\":2}",
+      "algorithm": "sha256",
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Approving. All five points from #588 are here, and the two that were actually open, the domain tag and the version discriminator, are inside the hashed bytes rather than beside them, which was the whole ruling.

**Verified the vectors independently rather than reading them.** Used the `rfc8785` package as an oracle with no cmcp code in the path, so the corpus is not checking itself:

- All five computable vectors reproduce exactly: `EAB-001`, `EAB-002`, `EAB-003`, `EAB-004`, `EAB-005`.
- The recorded `jcs_utf8` bytes match a fresh canonicalization byte for byte in every case.
- `EAB-006` is the refusal case and `sha512` is rejected as specified.
- Domain separation and version separation both change the digest, so those two vectors demonstrate the property rather than asserting it.

Fixing the expected JCS bytes in the file, alongside the digests, is what makes this reproducible by someone who disagrees with our implementation. That is the right shape for a contract vector set.

**One gap, recorded rather than blocking: the corpus cannot discriminate the ordering rule.** There is no character above U+FFFF anywhere in the file, and for all six vectors RFC 8785's output is byte-identical to a naive `json.dumps(sort_keys=True)`. `工具` in `EAB-002` is BMP, so it exercises multibyte UTF-8 encoding but not the place where JCS and a code-point sort diverge, which is UTF-16 code-unit ordering across the surrogate boundary. A verifier implementing the wrong ordering passes all six of these.

That is an addition rather than a correction, and it is the same gap measured in `trace-spec` last week, where 23 of 24 delegation-link vectors were sub-BMP and therefore satisfied by a correct canonicalizer and an incorrect one alike. Filed separately so it does not hold this.

Scope discipline is right too: docs and vectors only, no runtime activation, no admission-path change, and the `1` versus `1.0` witness explicitly retained as a governing-semantics question rather than treated as proof the binding must distinguish them.

PR #628, opened 2026-09-10
fix(session): hold the session boundary a credentialed reset draws

imran-siddique:
Found while reviewing the Monotonic Session Sensitivity State patent draft against the runtime. A credentialed reset is supposed to close one session and open a successor at the minimum sensitivity level. Four things let that boundary leak.

## A response arriving during a reset raised the successor

The per-session mutation lock serialised the reset and the response elevation but never ordered them, so whichever coroutine acquired it second won. A response in flight when the reset landed was applied to the successor, which had just been initialised to `public`, and recorded its pre-reset `call_id` as the call that raised it.

```
guard=False  successor max_sensitivity=confidential applied=True
guard=True   successor max_sensitivity=public       applied=False
```

`update_from_inspection()` now takes the `reset_count` observed at call entry and drops a response whose generation no longer matches, logging `SESSION_RESET_RACE`.

The discriminator is `reset_count` and not `session_id` because `upgrade_attestation()` rotates the identifier while deliberately continuing the same session at its current sensitivity, so a call in flight across an attestation upgrade must still apply. Both cases are tested.

The previous concurrency test asserted only that `max_sensitivity` remained a member of `SENSITIVITY_ORDER`, which every value satisfies.

## The reset route accepted the tool-invocation token

`POST /sessions/{id}/reset` is not reachable as an MCP tool, but it sat behind the same single `CMCP_BEARER_TOKEN` as `POST /mcp`, so an agent host holding only its own tool-invocation credential could clear accumulated session sensitivity.

The operator interface, session reset and catalog exception, now takes `CMCP_OPERATOR_TOKEN`. It must differ from `CMCP_BEARER_TOKEN` and is required outside `CMCP_DEV_MODE=1` (`OPERATOR_TOKEN_REQUIRED`). Where it is unset those routes still fall back to the bearer token, so an existing single-token deployment keeps working until it sets the new variable.

**Deployment note:** production deployments need `CMCP_OPERATOR_TOKEN` set before this ships, or startup aborts. Documented in `docs/configuration.md`.

## The reset audit entry did not identify the boundary

It recorded the sensitivity transition and nothing else. `detail` now carries the closed session identifier, the successor identifier, the resulting reset counter, and which credential was verified. `detail` is inside `_canonical_body()`, so those fields are covered by the entry hash.

## The chain attributed post-reset entries to the closed session

`AuditChain` stamps its construction-time `session_id`, so every entry written after a reset carried the closed session's identifier and the successor's identifier appeared nowhere in the chain. `rotate_session_id()` moves attribution to the successor after the boundary entry is written, so the reset entry belongs to the session that reached the recorded value.

The closed session's final value is also preserved now as a distinct `ClosedSessionRecord` rather than overwritten, and the reset response returns `closed_session_max_sensitivity` and `reset_count`.

## Testing

1625 passed, 9 new. Ruff and mypy clean on `src/`.

Two failures are pre-existing on the base branch, confirmed by stashing: `test_release_distribution_smoke` (runtime 0.5.0 against metadata 0.4.0) and `test_tpm_chained_verify` (assertion text drift).

https://claude.ai/code/session_01CDzMWcn12dszchgrKfXzZY


Qiang-Xu (approved):
LGTM.

imran-siddique:
Superseded by #629, which cherry-picks this onto current main and adds the shared, persistent session-state store.

PR #632, opened 2026-09-11
fix(trace): add missing agent_identity fields to the trace claim schema

### LINKED ISSUE #622: TRACE-001 is violated: claims carrying gateway.agent_identity fail the published schema
## cmcp-gateway version

`cmcp-runtime` 0.5.0, git `5b63642`

## Python version

3.12.10

## TEE provider

none (software-only claim, so the defect is independent of the provider)

## Reproduction steps

1. Build a claim exactly as `tests/unit/test_trace_claim.py::_make_claim()` does, which validates today.
2. Pass an `agent_identity` to `generate_trace_claim`, populated the way `session/manager.py:388` populates it in a deployment with an Agent Manifest binding, so `enforcement_mode` is set (`startup.py:638` always supplies it and the config default is `enforcing`).
3. Validate the result: `jsonschema.validate(instance=_to_dict(claim), schema=schemas/trace-claim.schema.json)`.

## Expected behavior

The claim validates. TRACE-001 in `tests/conformance/README.md` is a MUST: "Every TRACE Claim validates against the JSON Schema." README line 223 and `docs/SPEC.md` line 131 both call this file "the normative schema", and `docs/spec-index.md` tells implementers to run `ajv validate -s schemas/trace-claim.schema.json -d your-trace-claim.json` against their own output, so this is the documented acceptance step for an external verifier.

## Actual behavior

Validation fails. `gateway.agent_identity` sets `additionalProperties: false` and lists eight properties, while `AgentIdentityOut` has eleven. The three extras reach the signed claim whenever they are set, because `_to_dict` is `model_dump(exclude_none=True)` and its output is what `canonical_json` signs.

Two of the three are reachable in a normal deployment:

- `enforcement_mode`, set for any deployment with a manifest binding. Added by #590 on 2026-08-29.
- `intent_hash`, set whenever the bound manifest declares an intent. Added by #499 on 2026-08-11.
- `agent_key_thumbprint` is always `None` today, so `exclude_none` drops it. Latent only.

The schema has not changed since #424 on 2026-07-28, and all three fields were added after that date.

Nothing in the suite catches this. `tests/unit/test_trace_claim.py:556` is the only test that validates against the schema, and `_make_claim()` builds a claim with no `gateway.agent_identity`, so the object in question is never validated. No test references TRACE-001. That test arrived with #324, which was itself titled "fix gateway field drift".

```
TEST  claim with no agent_identity (the shape the suite validates)
  VALID against published schema

TEST  agent_identity carrying enforcement_mode (production shape)
  REJECTED at gateway/agent_identity
  Additional properties are not allowed ('enforcement_mode' was unexpected)

TEST  agent_identity carrying intent_hash
  REJECTED at gateway/agent_identity
  Additional properties are not allowed ('intent_hash' was unexpected)

TEST  agent_identity with only the eight documented fields
  VALID against published schema

TEST  is the field inside the signed pre-image?
  emitted keys: agent_id, authenticated_subject, enforcement_mode, issuer,
                issuer_key_id, manifest_id, policy_bundle_hash,
                subject_source, tool_catalog_hash
  enforcement_mode present in signed canonical JSON: True

pytest tests/unit/test_trace_claim.py -q  ->  34 passed

the documented step from docs/spec-index.md, against a signed claim
written to your-trace-claim.json by the runtime itself:
  INVALID at gateway/agent_identity
  Additional properties are not allowed ('enforcement_mode' was unexpected)
```

TRACE-001

## Suggested fix

1. Add the three properties to `gateway.agent_identity` in the schema, with patterns matching the `AgentIdentityOut` constraints.
2. Extend the schema conformance test to a claim carrying `agent_identity` with every optional field populated, so `additionalProperties: false` cannot drift again unnoticed, and reference TRACE-001 in its docstring.

One decision is needed before 1 can be written. `trace.policy.enforcement_mode` is schema-constrained to `enforce | advisory | silent` and `trace_claim.py:381` translates the runtime name into it, while `gateway.agent_identity.enforcement_mode` is an unconstrained `str | None` that `session/manager.py:397` fills with the runtime value untranslated. For a runtime configured `enforcing`, one signed claim therefore reads `enforce` in one field and `enforcing` in the other. Which vocabulary that field carries should be settled before it gets a schema entry, since typing it as a plain string would write the current inconsistency into the contract.

Happy to carry the PR once that is decided.

## Separately

`trace_claim.py:384` reads `mode_map.get(bundle.enforcement_mode, "advisory")`, so an unrecognised mode becomes `advisory` in a signed claim. `config.py:362` rejects anything outside `EnforcementMode`, so this is unreachable through configuration and latent rather than live, but `PolicyBundleInfo.enforcement_mode` is a plain `str` and is constructed directly under `experiments/` and in tests. A signed claim should not be able to assert a mode nobody configured, so raising is better than substituting.

The related question of where the correspondence between `silent` and the Agent Manifest's `audit-only` is written down is being handled in agentrust-io/agent-manifest#402.

qubeena07:
Closes #622.

gateway.agent_identity in schemas/trace-claim.schema.json declared 8 properties, while AgentIdentityOut (src/cmcp_runtime/audit/trace_claim.py:156) carries 11. The three missing ones, intent_hash, agent_key_thumbprint, and enforcement_mode, reach the signed claim whenever an Agent Manifest binding is configured. Since enforcement_mode defaults to enforcing and startup.py always supplies it, any deployment with a manifest binding was signing a claim that failed the project's own normative schema, a TRACE 001 violation.

imran-siddique confirmed the exact numbers and ruled on direction in the issue thread: fix the schema, not the code, since these three fields are real evidence rather than incidental plumbing. He also asked for a second part, a stated policy for what a verifier does when it meets claim members it does not recognize, since additionalProperties false on an evidence format that grows fields over time is the same problem tracked at agentrust io/trace spec#116, evidence outliving the verifiers that check it.

What this PR does

Adds intent_hash, agent_key_thumbprint, and enforcement_mode to gateway.agent_identity in the schema, matching the patterns and enum already used on the model and its sibling fields.

Adds a schema evolution paragraph to SPEC.md stating that additionalProperties false is a producer side authoring aid, not a contract external verifiers should enforce against a pinned older copy of the schema, and that a verifier should not reject a claim solely for carrying members it does not yet know. Cites trace spec#116 and names this issue as its first concrete instance in cmcp.

Updates the agent_identity field table and prose in docs/spec/session-policy.md so the docs describe all three fields, not just agent_key_thumbprint.

Adds a regression test in tests/unit/test_trace_claim.py that builds a claim with intent_hash and enforcement_mode set through generate_trace_claim and validates it against the schema. Confirmed it fails on the schema as it stood before this change and passes after.

Testing

Ran the full test suite, 1736 passed, 14 skipped. Three failures are present on main as well, unrelated to this change, all subprocess module resolution issues in a local editable install rather than anything to do with the schema or trace claim code. Ran ruff on the touched files, clean.

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -190,6 +190,8 @@ Verification (no operator trust required):
 4. Check `gateway.catalog.hash` against the approved catalog hash on record
 5. Recompute `trace.tool_transcript.hash` and walk `gateway.audit_chain` root to tip for call-level detail
 
+**Schema evolution.** `schemas/trace-claim.schema.json` is versioned by `cmcp_version`, and objects under it, `gateway.agent_identity` included, gain new optional properties across minor updates as the evidence the runtime signs grows (issue #622 was one such addition). `additionalProperties: false` in this file is a producer-side authoring aid, catching typos and unintended fields in what cMCP itself emits, not a contract that external verifiers should replicate against a pinned, possibly-older copy of the schema. A verifier MUST NOT reject an otherwise-valid claim solely because it carries object members the verifier's copy of the schema does not yet declare; it SHOULD instead validate the members it recognizes and ignore the rest. This is the same problem tracked more generally at `agentrust-io/trace-spec#116`, evidence outliving the verifiers that check it, and #622 is its first concrete instance in cMCP.
+
 ---
 
 ## 6. Attestation Bindings (Phase 1 unique properties)
--- a/docs/spec/session-policy.md
+++ b/docs/spec/session-policy.md
@@ -194,6 +194,8 @@ This binding answers "who acted" for the session. It does not replace `trace.sub
 
 Offline verifiers SHOULD cross-check `gateway.agent_identity` against the signed manifest and trusted issuer key. This keeps the runtime boundary check and the evidence artifact self-checking.
 
+`gateway.agent_identity` MAY also carry `intent_hash` (AARM R2: a digest of the issuer-signed declared intent, agent-manifest spec §3.9) and `enforcement_mode` (the Agent Manifest binding's enforcement mode at session creation, distinct from `trace.policy.enforcement_mode`). Both are populated whenever the bound manifest supplies them, which is the common case once `agent_manifest` is configured.
+
 `gateway.agent_identity` MAY also carry `agent_key_thumbprint`: an RFC 7638 JWK thumbprint of the agent's own
 signing key, rendered as `sha256:<hex>`, distinct from `issuer_key_id` (the key that signed the manifest).
 It is optional and is omitted from every claim the current runtime can produce, since no code path here has
@@ -210,4 +212,4 @@ The following fields from session state are included in the TRACE attestation re
 |-------|------|-------------|
 | `session_max_sensitivity` | string | The highest `max_sensitivity` value reached during the session. |
 | `session_reset_count` | integer | Number of times `POST /session/reset` was called during the session lifetime. Normally `0`; a non-zero value warrants review. |
-| `agent_identity` | object | Optional Agent Manifest binding: manifest ID, bound agent ID, authenticated subject, subject source, issuer key ID, policy hash, catalog hash, and an optional `agent_key_thumbprint`. Present only when `agent_manifest` is configured and verified. |
+| `agent_identity` | object | Optional Agent Manifest binding: manifest ID, bound agent ID, authenticated subject, subject source, issuer key ID, policy hash, catalog hash, and the optional `intent_hash`, `agent_key_thumbprint`, and `enforcement_mode`. Present only when `agent_manifest` is configured and verified. |
--- a/schemas/trace-claim.schema.json
+++ b/schemas/trace-claim.schema.json
@@ -235,6 +235,21 @@
             "tool_catalog_hash": {
               "type": "string",
               "pattern": "^sha(256|384):[0-9a-f]+"
+            },
+            "intent_hash": {
+              "type": "string",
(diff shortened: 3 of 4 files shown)


rajnisht7:
the regression test `test_claim_schema_accepts_current_optional_fields` covers `intent_hash and `enforcement_mode`, but agent_key_thumbprint` is not actually set in the test. As this change adds all three fields, can you also populate agent_key_thumbprint in the test

imran-siddique (approved):
Approving. Both halves of the #622 ruling are here: the three properties in `schemas/trace-claim.schema.json`, and the statement in `docs/spec/session-policy.md` that a verifier MUST NOT reject a claim solely for carrying members its own copy of the schema does not declare, with trace-spec#116 named as the general case. The second half is the part that stops this recurring for whoever pinned the old file.

@rajnisht7 was right that `agent_key_thumbprint` had no coverage, and I have pushed it as `77e94d0` rather than send this back for one line.

It needed a sibling test rather than an addition to the existing one, and the reason is worth recording. `session/manager.py:385` omits the thumbprint deliberately, because `AgentManifestBinding` carries no agent key bytes yet, and the schema description says it is only ever set when `subject_source` names a live-authenticated source. The existing case uses `subject_source="config"`, so setting the thumbprint there would have pinned a combination this runtime is documented never to emit. The new test uses `subject_source="svid"`, which is the shape the schema has to accept on the day #425 wires a real key source.

Verified in both directions: it fails against `main`'s schema and passes against this branch's. Full CI green on `77e94d0`, including all six test matrix jobs and `governance`.

Merging.

Issue #566, opened 2026-08-24
Bind progressive MCP discovery to versioned catalog checkpoints


imran-siddique:
Support progressive discovery and server/discover without weakening catalog approval. Each governed call must bind to the exact discovered schema and approved catalog checkpoint. Acceptance evidence: cache refresh, stale catalog, tool removal, schema substitution, and unapproved late-discovery tests.

solloek369-arch:
Hi,
I’ve been looking at #566 from a conformance angle and did a read-only pass over the current call/catalog path. There are already useful partial controls for catalog membership, advertised-vs-approved schema identity, and catalog drift, but the integrated #566 discovery/checkpoint/call-basis path is not built yet.

For that future path, one fixture shape may be useful: keep three identities independently mutable — the discovered schema, the approved catalog checkpoint, and the checkpoint attributed to a governed call.

That gives three focused differentials: unapproved late discovery, schema-only substitution under the same apparent tool identity, and a historical call being rebound to a later checkpoint.

The aim is not to change the restart/immutability model, only to make those identities independently testable as #566 lands. Would that fixture shape add useful coverage to the acceptance matrix?

imran-siddique:
The discovered schema, approved checkpoint, and call-attributed checkpoint need independent acceptance cases. Keep the suggested conformance matrix here; implementation ownership is not established by a review comment alone. Coordinate the manifest side with agentrust-io/agent-manifest#340.

solloek369-arch:
Building on the three independently mutable identities already retained here, I think the acceptance matrix can add two things without repeating the existing fixture proposal:

1. **Make the discovery→checkpoint pipeline explicit.** Separate acquisition, structural validation, projection, and comparison. MCP `tools/list` is paginated, while current cMCP `5b63642` consumes one returned `tools` array in this path; discovery also collapses tools by name, and the portable checkpoint byte recipe remains a design input. Could the matrix require pagination exhaustion with repeated/cyclic-cursor controls, refusal of duplicate or malformed discovery input before projection, and one pinned canonical byte recipe before comparison?

2. **Add a legitimate-advance positive control.** Alongside the existing late-discovery, schema-substitution, and historical-rebinding negatives, include C0→C1 where a historical call remains correctly attributed to approved C0 and a later call uses separately approved C1. That distinguishes correct evolution from an implementation that passes by rejecting every change. It does not imply C0 remains authorized for new calls, or that historical transition provenance alone establishes current authorization.

[...]

PR #633, opened 2026-09-13
fix(discovery): exhaust paginated tools/list before comparison

### LINKED ISSUE #631: tools/list pagination is never exhausted: a multi-page catalog is read as a changed catalog
`tools/list` is paginated in MCP and cMCP consumes one page, so a server with more tools than fit in a single response gives the gateway a partial catalog. Raised by @solloek369-arch on #566 and verified on `main`.

## The evidence

`src/cmcp_runtime/mcp/proxy.py`, `_upstream_tools`:

```python
payload, headers = build_request("provenance-tools-list", "tools/list", {})
resp = await client.post(entry.server.url, json=payload, headers=headers)
resp.raise_for_status()
result = parse_response(resp, "provenance-tools-list").get("result")
...
tools = result.get("tools") if isinstance(result, dict) else None
return tools if isinstance(tools, list) else None
```

Empty params, so no cursor is sent. One read of `result["tools"]`, and return.

The string `cursor` appears **zero times in the entire `src/` tree**, so this is not a gap in one path. There is no pagination handling anywhere in the codebase, and `nextCursor` is discarded silently because nothing looks for it.

## Why it matters here specifically

`_check_upstream_drift` compares what a server advertises against what was approved. With a truncated advertisement, that comparison runs against a catalog that is missing tools rather than one that has changed.

The failure direction is the bad one. A tool present in the approval and absent from page one looks like a removal; a tool added on page two is invisible entirely. Neither reads as "we did not finish reading the catalog", which is what actually happened.

## What is needed

1. Exhaust pagination: follow `nextCursor` until the server stops returning one.
2. Guard the loop: a repeated cursor and a cyclic cursor must both terminate and be reported as a discovery failure rather than as a completed read.
3. Report an incomplete read as incomplete. A catalog that could not be fully acquired is not a catalog that changed, and the drift comparison must not run on one.

Point 3 is the one that matters beyond correctness: an unfinished acquisition currently becomes a drift verdict, which is a not-established case being reported as a result.

## Scope

Discovery and drift only. No change to approval semantics, checkpoint binding, or the acceptance matrix being designed on #566; the matrix will want pagination-exhaustion cases once this exists.

Related: #566, #568.

noah-ing:
## What

Exhaust upstream `tools/list` pagination before drift or provenance comparison, using one acquisition helper for HTTP and stdio.

- Pass opaque cursors back unchanged, including the empty string; only an absent `nextCursor` completes discovery.
- Discard the whole acquisition on a later-page error, malformed response, duplicate tool name, repeated/cyclic cursor, or continuation beyond 1,000 pages.
- Preserve request-scoped HTTP JSON/SSE handling and metadata. Cancellation propagates without caching an unfinished acquisition/check, and closes an interrupted stdio child so a late reply cannot contaminate a later call.
- Share one completed first-contact acquisition between drift and provenance for the same server/publisher authority in a session. Concurrent readers wait for completion. Session close/reset drains admitted calls before invalidating acquisition and comparison caches together.

## Why

Closes #631. Credit to @solloek369-arch for the observation on #566 and @imran-siddique for confirming and filing the issue.

An approved tool on page two previously looked withdrawn; conversely, a matching first page could conceal failed acquisition of a later page. The initial transport regressions on unchanged production code produced **37 failures / 29 existing passes**.

The [duplicate-acquisition cost identified by qubeena07](https://github.com/agentrust-io/cmcp/pull/633#issuecomment-5654294598) is addressed in `8b8ba3b`. Twelve additional public-call transport regressions fail on the preceding commit with **four discovery requests instead of two**, then pass with the shared cache. Both comparisons still run against the completed listing.

## Security impact

Incomplete acquisition is `unchecked`, never a comparison against partial tools. The existing behavior allowing calls when discovery is unchecked remains unchanged; this PR does **not** turn that state into a fail-closed authorization gate. A configured required-provenance floor still prevents forwarding when provenance is unchecked.

No changes to approval semantics, catalog hash construction, checkpoint binding, signature behavior, or #566 acceptance-matrix redesign. The page limit bounds request count, not total elapsed time or bytes. Pagination does not provide an atomic snapshot of a changing server. Validation covers acquisition shapes and unambiguous names, not the full MCP tool schema. First-contact caching is not continuous monitoring: completed unchecked acquisitions are cached too. Cancellation is not cached, but the existing stdio child-close behavior does not automatically restart that child. These limits are recorded in `LIMITATIONS.md`.

## Integration with #634

Merged current `main` (`c9d0133`) into this branch at `a9be23a`. The lifecycle's admission/drain ordering and retryable resource cleanup are preserved. Cleanup invalidates all four discovery/comparison cache containers before any cleanup await or empty-resource return; a failed child close does not adopt the successor.

Eight additional cases cover normal drain and owner/waiter cancellation across HTTP JSON, SSE, and real stdio, plus empty/failed cleanup. They assert old-call finalization before rebind, fresh successor resources and discovery, signed provenance, and separated audit records. The six transport/lifecycle cases also passed ten consecutive repetitions. The defensive cache-generation unit test is explicitly separate from the public lifecycle: rebinding while public calls remain active is not permitted.

Windows CI on `a9be23a` found a test-teardown socket reset after the stdio child was deliberately terminated. `1a01914` narrowly handles that expected reset at `wait_closed()`; the authorization, ordering, cancellation, and resource assertions are unchanged. Full host and clean no-cache container checks were rerun on this final head. [Final hosted CI](https://github.com/agentrust-io/cmcp/actions/runs/34789375673) passed all six Python 3.11–3.13 / Ubuntu and Windows test jobs plus governance. CodeQL, changed-code fuzzing, Docker, and limitations parity passed; the conditional benchmark was skipped. Current-head maintainer approval remains required.

## Test plan — integrated head `1a01914`

- [x] Hosted matrix: **1,845 passed / 6 existing skips** per Ubuntu job; **1,844 passed / 7 existing skips** per Windows job. All six test jobs passed.
- [x] Full host suite: **1,845 passed / 6 existing optional or hardware-fixture skips**, 88.59% coverage.
- [x] Clean committed-source, no-cache Linux container: **1,845 passed / 6 existing skips**, 88.63% coverage, plus Ruff, mypy, Bandit, and `pip check`; runtime network disabled, non-root, all capabilities dropped.
- [x] `ruff check src/ tests/` and `mypy src/cmcp_runtime/ src/cmcp_verify/` pass. Ruff formatting passes for the new test files and modified production-code range; unrelated pre-existing formatting is unchanged.
- [x] **113 added cases overall**, including HTTP JSON/SSE, real stdio subprocess I/O, opaque/empty cursors, exact page-budget boundary, cycles, late failures, actual drift under both policies, signed provenance outcomes, and cancellation cleanup. Existing withdrawal tests remain passing. No new skip/xfail.
- [x] **25 cache-follow-up cases** cover full cold/warm tool calls, signed provenance, required-provenance denials, completed empty/unchecked results, concurrent acquisition/comparison, cancellation ownership, identity isolation, independent-server progress, and session-boundary invalidation and defensive cache-generation handling. The new #634 integration cases are additional to these 25.
- [x] Provenance cases use actual TRACE signing/verification: complete match → `VERIFIED`, complete change → `CATALOG_MISMATCH`, incomplete discovery → `UNCHECKED`.
- [x] Bandit, dependency audit (`--skip-editable`), Gitleaks diff scan, detect-secrets changed-file scan, and whitespace checks pass.
- [x] Wheel/sdist build and Twine validation; each artifact installed and smoke-tested outside the checkout in a fresh environment, including packaged code/schema resources and CLI availability.
- [x] Separate read-only source review found no concrete integration defect; changelog includes original-reporter and follow-up-reviewer attribution.

Host and container use the repository's hash-locked `requirements/dev.txt`; no dependencies were added or changed. These are software protocol regressions, not hardware-attestation evidence.

## DCO sign-off

- [x] I certify that I wrote or have the right to submit this contribution, and I agree to the [Developer Certificate of Origin](https://developercertificate.org/). Commit signed off as Noah Ingwers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Exhaust upstream `tools/list` pagination before drift and provenance
+  comparisons** (#631). HTTP and stdio share bounded acquisition: later-page
+  failures, malformed or ambiguous listings, and cursor cycles are unchecked,
+  never a comparison against a partial catalog. Existing drift policy and
+  unchecked-call behavior are unchanged. Observed by solloek369-arch on #566
+  and confirmed in #631 by Imran Siddique.
+
 - TLS pinning test fixtures set `minimum_version = TLSv1_2`; the server was built
   with `PROTOCOL_TLS_SERVER` and no floor, leaving TLSv1 and TLSv1.1 reachable in
   the test that asserts the gateway's transport rules.
--- a/LIMITATIONS.md
+++ b/LIMITATIONS.md
@@ -25,6 +25,8 @@ cMCP compares what each upstream server advertises against the approved catalog 
 
 Separately, the approved description rather than the live one is what the gateway serves to the agent on `tools/list`, so a mutated description does not reach the model through cMCP even in the windows above. That is a structural property of proxying an approved catalog, not a detection result, and it does not extend to the tool's behaviour once called.
 
+HTTP and stdio discovery exhaust `tools/list` pagination before comparing either drift or provenance. A later-page failure, malformed discovery shape, duplicate tool name, repeated/cyclic cursor, or continuation beyond 1,000 pages makes the entire acquisition unchecked; no partial list is compared. Cursors are passed back unchanged, including an empty string. This bound limits page count, not total elapsed time or response bytes, and pagination does not establish an atomic snapshot of a changing server. This is acquisition validation, not full MCP schema validation or a new approval/hash policy. The unchecked-call behavior above is unchanged.
+
 **Phase 2 completeness: server-side attestation**
 Phase 1 attests the gateway boundary. It does not attest what happens on the other side of that boundary. The `tool_transcript.hash` field in the TRACE Claim records a hash of the audit chain tip, but the tool transcript binding that ties a specific tool execution to a specific response is Phase 2 work. Phase 1 partially addresses P1.4 (transitive trust into upstream dependencies) and P4.1 (typosquatted packages added to catalog) -- both are fully closed by Phase 2. Any compliance claim that relies on server-side proof must wait for Phase 2.
 
--- a/src/cmcp_runtime/mcp/discovery.py
+++ b/src/cmcp_runtime/mcp/discovery.py
@@ -1,0 +1,68 @@
+"""Acquire a complete tools/list before drift or provenance comparison.
+
+This is acquisition validation, not approval or a catalog hash construction.
+No partial result escapes: every page must be attributable and well-shaped,
+and pagination must terminate within the local page budget.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Awaitable, Callable
+from typing import Any
+
+# A page bound also terminates servers that issue endlessly distinct cursors.
+# It is not a wall-clock deadline or a guarantee of an atomic server snapshot.
+MAX_DISCOVERY_PAGES = 1000
+
+PageFetcher = Callable[[str, dict[str, Any]], Awaitable[dict[str, Any]]]
+
+
+class DiscoveryError(ValueError):
+    """A bounded local reason; never embed upstream payloads or cursor values."""
+
+
+async def collect_tools(fetch_page: PageFetcher) -> list[dict[str, Any]]:
+    """Return only an exhausted, unambiguous listing; otherwise raise.
+
+    Transport failures and cancellation propagate to the owning transport.
+    Cursors are opaque strings, including the empty string. Only absence of
+    nextCursor terminates a listing; an empty page alone does not.
+    """
+    tools: list[dict[str, Any]] = []
+    names: set[str] = set()
+    cursors: set[str] = set()
+    params: dict[str, Any] = {}
+    for page in range(MAX_DISCOVERY_PAGES):
+        request_id = f"provenance-tools-list-{page}"
+        body = await fetch_page(request_id, params)
+        if (
+            not isinstance(body, dict)
+            or body.get("jsonrpc") != "2.0"
+            or body.get("id") != request_id
+            or "error" in body
+        ):
+            raise DiscoveryError("invalid response envelope")
+        result = body.get("result")
+        if not isinstance(result, dict) or not isinstance(result.get("tools"), list):
+            raise DiscoveryError("invalid tools page")
+        for tool in result["tools"]:
+            if (
+                not isinstance(tool, dict)
+                or not isinstance(tool.get("name"), str)
+                or not tool["name"]
+            ):
+                raise DiscoveryError("invalid tool name")
+            if tool["name"] in names:
+                raise DiscoveryError("duplicate tool name")
+            names.add(tool["name"])
+            tools.append(tool)
+        if "nextCursor" not in result:
+            return tools
+        cursor = result["nextCursor"]
+        if not isinstance(cursor, str):
+            raise DiscoveryError("invalid continuation cursor")
+        if cursor in cursors:
+            raise DiscoveryError("repeated continuation cursor")
+        cursors.add(cursor)
+        params = {"cursor": cursor}
+    raise DiscoveryError("discovery page limit exceeded")
--- a/src/cmcp_runtime/mcp/proxy.py
+++ b/src/cmcp_runtime/mcp/proxy.py
@@ -34,6 +34,7 @@ from cmcp_runtime.config import Config, DriftPolicy
 from cmcp_runtime.errors import PolicyDeny, UpstreamToolError, UpstreamUnavailable
 from cmcp_runtime.execution import valid_execution_id
 from cmcp_runtime.mcp import tls_pinning
+from cmcp_runtime.mcp.discovery import DiscoveryError, collect_tools
 from cmcp_runtime.mcp.stdio import StdioServer
 from cmcp_runtime.mcp.streamable_http import (
     build_request,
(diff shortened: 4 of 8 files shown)


qubeena07:
Reviewed the pagination exhaustion logic in discovery.py and its integration into proxy.py and stdio.py, cursor handling including the empty string vs absent distinction, duplicate and cycle detection, the page budget boundary, and stdio cancellation cleanup. All of that checks out and the 79 new tests cover it well.

One thing worth flagging, not blocking, more a heads up on cost. _check_upstream_drift and _check_provenance each call _advertised_tools independently, with separate first contact caches, self._drift_checked and self._provenance. On the very first tool call against a server that has provenance_record_path configured, both caches are empty, so _call_tool_impl calls _check_upstream_drift directly, then later in the same call _forward_to_upstream calls _check_provenance, and both trigger the full paginated tools/list fetch.

Before this PR that was cheap either way, one request per call. Now _advertised_tools can be up to MAX_DISCOVERY_PAGES round trips, and cold start against a server with provenance configured just paid that cost twice back to back instead of once, doubling stdio child process round trips too on that path.

Given the PR's own note that the page limit bounds request count rather than elapsed time, this seems worth closing, maybe by caching the _advertised_tools() result itself per server per session rather than caching only the two downstream comparisons, so both callers share one fetch. Happy to see this land separately if you would rather not hold the PR on it.

noah-ing:
Thanks for tracing the cold-call path — you're right. Fixed in 8b8ba3b.

Drift and provenance now share one completed paginated acquisition per server/publisher authority per session. Concurrent readers wait for that acquisition, and a drift check is not marked complete while discovery is still in flight. Completed unchecked outcomes are cached distinctly from empty catalogs; cancellation is not cached. Session rebinding resets the acquisition and both comparison caches together, with old in-flight results discarded before reuse.

Added 25 regressions. The 12 public-call transport cases use HTTP JSON/SSE or real stdio subprocess I/O and actual signed TRACE provenance verification. All 12 fail on the preceding commit with four discovery requests instead of two, then pass with this change. They also cover warm reuse, warn-only mismatch, and incomplete discovery with and without a required-provenance floor. The other cases cover concurrency, cancellation ownership, identity isolation, and session reset.

Full host and clean no-cache Linux runs each pass **1,803 tests, with six unchanged optional/hardware skips**. Ruff, mypy, Bandit, dependency/secret scans, and fresh wheel/sdist installation checks pass; all 29 newly changed executable lines are covered.

The existing unchecked-call policy is unchanged, as are the page-count versus elapsed-time limit and non-atomic-snapshot caveats. The changelog credits your observation.

carloshvp (dismissed):
Reviewed exact head 8b8ba3b3dba9d3816553d1267f1600224aaeb339. No blocking findings in the pagination and discovery-cache change.

Complete acquisition precedes drift/provenance comparison on HTTP JSON/SSE and stdio. Later-page failures, invalid envelopes, duplicate names, cursor cycles, and over-budget continuations do not yield a partial catalog. Concurrent checks share completed acquisition; cancellation is not cached, and rebinding discards old acquisition/cache state. The existing unchecked-discovery policy remains separate from the configured required-provenance floor.

Independent isolated macOS/Python 3.13 validation: full suite 1,845 passed, 14 skipped; Ruff, mypy, Bandit and whitespace checks passed. An additional seeded matrix exercised 300 multi-page acquisitions, including 100 late duplicate failures, plus the actual 1,000-page completion/continuation boundary. Replacing the changed proxy/stdio implementations with main's versions makes the relevant transport suites fail: 62 failed, 29 passed. Current tests include signed TRACE provenance and real stdio subprocess transport, not only asserted verification results.

The merge simulation against freshly fetched main is conflict-free and all four commits have DCO sign-offs. #634 touches the same session/resource lifecycle code and will need careful integration and fresh combined validation when the second PR lands. This approval covers this head independently; remaining repository gates and platform/hardware skips are not cleared by local testing.

carloshvp (approved):
Re-reviewed exact head 1a0191477d741492f9a9def800271d5dc4189be0 after integration with merged #634. No blocking findings.

The lifecycle integration preserves admission/drain/retry behavior while invalidating the advertised catalog, discovery locks, provenance, and drift caches together before resource cleanup. The new public-call tests exercise normal drain and deadline cancellation through HTTP JSON, SSE, and real stdio with signed TRACE provenance. The Windows socket-reset adjustment is confined to expected teardown errors in the test gate.

Independent isolated macOS/Python 3.13 validation: full suite 1,887 passed, 14 skipped; Ruff, mypy, Bandit and whitespace checks passed. Three additional real-subprocess endpoint checks verify that close, reset and shutdown clear the paginated acquisition cache and locks, terminate the old child, and (for close/reset) reacquire both pages in a fresh successor. Those checks were mutation-tested against the earlier cleanup behavior: all three fail if cleanup clears only provenance/drift and retains the new acquisition caches.

I had prepared a local conflict resolution, but the final pre-push check found your newer commits. I did not push over them. Their runtime integration matches the independently validated fix; the results above are from your exact new head. Current hosted technical checks pass, the fetched-main merge simulation is clean, and the commits are signed off. Approving this head; hardware/platform skips remain outside local validation.

imran-siddique (approved):
Approving. Read the acquisition loop rather than the summary, because the property #631 is about is the one that is easy to implement almost-correctly.

It terminates on `"nextCursor" not in result` alone, so an empty-string cursor continues the listing. That is the case the MCP spec makes explicit and the one a truthiness check gets wrong, and it is the difference between exhausting a catalog and stopping one page early on a server that pages with `""`.

The discard paths are the other half and they are all present: a malformed envelope, a malformed page, a non-string or empty tool name, a duplicate tool name, a non-string cursor, a repeated cursor, and the page ceiling each raise rather than return what was collected so far. A partial acquisition never reaches the drift or provenance comparison, which is what #631 was actually about: the comparison was being fed a truncated catalog and reporting agreement.

141 passed and 1 skipped across the six discovery test files here.

Merging.

PR #652, opened 2026-09-15
docs(execution-action-binding): record that the vectors cannot test member ordering

### LINKED ISSUE #630: execution-action-binding v1 vectors cannot discriminate RFC 8785 ordering from a code-point sort
The execution-action-binding v1 vector set merged in #627 cannot distinguish a conformant RFC 8785 canonicalizer from one that sorts by code point. Measured on the merged file.

## The measurement

`docs/spec/execution-action-binding-v1-vectors.json` contains six vectors. Using the `rfc8785` package as an oracle:

- no character above U+FFFF appears anywhere in the file;
- for all six vectors, `rfc8785.dumps(preimage)` is byte-identical to `json.dumps(preimage, sort_keys=True, separators=(',',':'), ensure_ascii=False)`.

So an implementation that sorts member names by Unicode code point, rather than by UTF-16 code unit as RFC 8785 Section 3.2.3 requires, reproduces every expected digest in the set and passes.

`EAB-002-nonascii-sha256` uses `工具` (U+5DE5, U+5177). That exercises multibyte UTF-8 encoding, which is worth having, and it is entirely inside the BMP, where the two orderings agree.

## Why the two orderings diverge, and where

RFC 8785 sorts by UTF-16 code unit. For characters above U+FFFF, UTF-16 encodes a surrogate pair beginning in `D800`-`DBFF`, which sorts *below* BMP characters in `E000`-`FFFF`. Sorting by code point puts them above. So the sets disagree only when a preimage carries both a supplementary-plane member name and a BMP member name in that range.

## What to add

One vector whose preimage has two member names chosen so the two rules produce different orderings, with the expected JCS bytes and digest fixed in the file as the existing six are. A conformant implementation reproduces it; a code-point sorter does not.

The binding contract's fields are fixed at `domain`, `version`, `agent_id`, `action_type`, `action_scope`, `action_timestamp`, all ASCII, so the discriminating case has to come from member names inside a value that is itself an object, or the vector set should state plainly that ordering is not exercised because the preimage shape makes it unreachable. Either answer is fine; the current position of appearing to test it while not testing it is the one to avoid.

## Not specific to this repository

The same measurement holds in `agentrust-io/trace-spec`: 23 of the 24 delegation-link vectors are sub-BMP, so the corpus is satisfied by a correct canonicalizer and an incorrect one alike. Raised there on trace-spec#231 on 10 September. Recording it here so the two are not solved twice or, worse, once.

Related: #588, #627, agentrust-io/trace-spec#231.

qubeena07:
Closes #630.

The six vectors in execution-action-binding-v1-vectors.json cannot tell an RFC 8785 conformant canonicalizer, which sorts object members by UTF 16 code unit, from one that sorts by raw Unicode code point instead. The two orders only disagree when an object has one member name in the supplementary plane, above U+FFFF, encoded in UTF 16 as a surrogate pair, alongside another member name in the E000 through FFFF range. None of the six vectors have that shape, so both a conformant and a non conformant canonicalizer reproduce every digest in the set and pass.

The issue named two acceptable fixes. Add a vector that actually discriminates the two orderings, or state plainly that ordering is not exercised because the preimage shape makes it structurally unreachable. Read the contract in execution-action-binding.md before picking one. The v1 preimage is fixed to exactly six ASCII named members, domain, version, agent_id, action_type, action_scope, action_timestamp, each holding a plain string or the integer 1, never a nested object, and the contract itself says a producer must not add, remove, rename, or reinterpret a member while still calling the result a v1 execution action binding. There is no way to introduce a member name pair that would diverge under the two sort rules without breaking that rule. So a real discriminating vector cannot be built for this profile, the second option is the correct fix, not a shortcut.

What this does

Adds a paragraph to the Conformance evidence section of execution-action-binding.md stating this plainly, with the reasoning above, and pointing at agentrust io slash trace spec issue 231 as the same structural gap already recorded in a different vector set.

Adds a matching short note as a top level field in execution-action-binding-v1-vectors.json itself, so someone reading only the JSON sees the same statement without needing the prose doc.

No code or tests touched. Nothing in this repo consumes this vectors file as a runtime fixture, it is a conformance artifact for external implementers, confirmed by grepping src and tests for any reference to it.

Testing

Ran the full suite, 1884 passed, 14 skipped, same 3 pre existing failures as on main, all unrelated local environment issues with subprocess module resolution, nothing to do with this change. Validated the edited JSON file parses cleanly.


imran-siddique (changes requested):
The fixed-field explanation addresses the documentation option in #630. Before merging, please correct two statements:

- Replace “sorting below the Basic Multilingual Plane” with “sorting before BMP characters in U+E000 through U+FFFF.” The ordering is defined in [RFC 8785 §3.2.3](https://www.rfc-editor.org/rfc/rfc8785.html#section-3.2.3).
- Remove the claim that TRACE #231 has the same structural limit. That thread [confirms a working discriminating vector](https://github.com/agentrust-io/trace-spec/issues/231#issuecomment-5641646011); its preimage permits member names that this profile excludes.

Commit `b737f08` also lacks the `Signed-off-by` trailer required by CONTRIBUTING.md. Please supply your own sign-off when updating the branch. Could you make these corrections by September 17?

qubeena07:
Pushed 01e41e0 with all three fixed.

Corrected the ordering description to sorting before BMP characters in U+E000 through U+FFFF, matching RFC 8785 section 3.2.3, instead of the vaguer Basic Multilingual Plane phrasing.

Removed the trace-spec#231 comparison entirely, since that thread actually has a working discriminating vector and its preimage allows member names this profile does not, so it is not the same structural limit.

Amended the commit with a sign off, git commit -s, per CONTRIBUTING.md.

imran-siddique (approved):
Re-reviewed 01e41e004c6612f14a5deec9f128f914de8cd704. All three requested corrections are present: the Unicode ordering statement names U+E000 through U+FFFF, the incorrect TRACE comparison is gone, and the contributor supplied their own DCO sign-off.

Recomputed the five supplied canonical byte strings and SHA-256/SHA-384 digests, checked the sixth vector's unsupported-algorithm disposition, and used a supplementary-plane/BMP key pair as a discriminating ordering control. The fixed ASCII member names cannot expose that ordering difference. No vector verdict or runtime behavior changes in this PR; hosted checks pass.

PR #656, opened 2026-09-17
Enforce verifier-owned SNP platform policy

imran-siddique:
## What

Add immutable `SnpPlatformPolicy` requirements to native SNP, Azure vTPM-rooted SNP and the public `verify_trace_claim` API. Explicit policy requires authenticated SNP evidence and is evaluated only after signature/chain verification.

## Why

A valid report signature establishes authenticity but does not establish that the reported platform meets the relying party's requirements. Callers can now require or forbid PLATFORM_INFO fields, reject unknown bits, and receive a failure when software evidence or another provider cannot satisfy the configured SNP policy.

## Security impact

Policy is supplied by the verifier, not accepted from the claim issuer. Unknown or contradictory field requirements are rejected. Omitting the optional policy preserves existing verification behavior.

This covers PLATFORM_INFO only. Guest DEBUG policy, firmware TCB currency, revocation, GPU state and runtime gateway admission remain outside this change. Synthetic signed reports test enforcement; this PR does not claim a new cMCP hardware run.

## Test plan

- [x] `pytest tests/unit tests/conformance tests/integration -q`: 1,871 passed, 8 skipped on Windows/Python 3.12
- [x] `ruff check src/ tests/`
- [x] `mypy src/cmcp_runtime/ src/cmcp_verify/`
- [x] `bandit -q -r src/ -c pyproject.toml`
- [ ] New manual cMCP hardware validation (not performed)

Includes 28 focused platform-policy tests using signed synthetic evidence, covering both SNP paths, the public API and provider/software bypass attempts. Checks ran on the branch rebased onto current main with matching local package metadata.

## DCO sign-off

- [x] Contribution is submitted with DCO sign-off.


Qiang-Xu (approved):
Looks good.

PR #661, opened 2026-09-18
fix(config,startup): accept and wire session_state_path (#653)

### LINKED ISSUE #653: session_state_path is documented and released but rejected at config load and never constructed at startup
## cmcp-gateway version

## Python version
N/A — defect found by source review, reproduced against 4ca6102

## TEE provider
N/A

## Reproduction steps
1. Add `session_state_path: /some/path/state.db` to cmcp-config.yaml and run `validate-config`.
2. Observe it aborts: the key is rejected before parsing.
3. Patch `_KNOWN_TOP_KEYS` to allow the key, then start the gateway.
4. Inspect the returned RuntimeContext: `session_state_store` is still None.
5. Start two instances over the same path — the session-sensitivity ratchet is not shared.

## Expected behavior
`session_state_path` is accepted and a shared, persistent `SqliteSessionStateStore` is constructed and passed into RuntimeContext, so the ratchet survives restart and holds across horizontally scaled instances, as documented and released.

## Actual behavior
Two independent breaks:
1. `_KNOWN_TOP_KEYS` (src/cmcp_runtime/config.py:158-171) omits `session_state_path`, so `load_config` rejects the documented key before the traversal/parse code at config.py:489-491 and :561 is ever reached.
2. Even with the key allowed, `run_startup` never constructs `SqliteSessionStateStore` (only an import + a dataclass annotation exist in src/). RuntimeContext keeps `session_state_store=None`, `SessionState.hydrate()` returns False, and every instance keeps a private in-process ratchet.

The store's own 10-test suite passes and the rest of the suite stays green, so CI does not catch the gap.

`validate-config` aborts on the documented key; after allowlisting, `ctx.session_state_store is None`.

## Additional context
This contradicts shipped docs and release notes:
- docs/configuration.md:124 promises the shared persistent SQLite store (BEGIN IMMEDIATE serialisation + path-traversal rejection).
- CHANGELOG.md:71-84 says the accumulated value "can now live in a shared, persistent store" and that the gateway hydrates at call entry.
- Merged PR #629 states the feature as implemented: https://github.com/agentrust-io/cmcp/pull/629
- The repo's own wiring doctrine (src/cmcp_runtime/cli.py:46-51): "All components validated by run_startup() MUST be wired here, a component that is validated but not passed through is silently inert in production (the AUTH-001 bearer token and AUDIT-001 store were both lost this way)."

Yatsuiii:
Closes #653. Credit to @saintmalik for the report - the root-cause tracing to both defects, including the exact `config.py`/`startup.py` line citations, is theirs; this PR is the fix.

`session_state_path` was documented (`docs/configuration.md`) and its feature described as shipped (CHANGELOG, merged #629), but was non-functional in two independent ways:

1. `_KNOWN_TOP_KEYS` (`config.py`) omitted the key, so `load_config` rejected any config that set it, before the existing parsing code (`config.py:489-491`, `:561`) ever ran.
2. Even past that, `run_startup` never constructed a `SqliteSessionStateStore` - only an import and a `RuntimeContext` dataclass field existed. `RuntimeContext.session_state_store` stayed `None` regardless of configuration, so the shared, persistent session-sensitivity ratchet never activated: a restart still lost the accumulated value, and multiple instances serving one session still ratcheted independently.

Confirmed both by direct inspection before writing anything: `SqliteSessionStateStore(` had zero call sites in `src/` outside its own class definition.

## What changed

- `config.py`: added `session_state_path` to `_KNOWN_TOP_KEYS`.
- `startup.py`: added `_open_session_state_store()`, called from `run_startup` as step 5g, after the kill switch store's 5f. Opens the store when `session_state_path` is configured and wires it into `RuntimeContext`; leaves it `None`, unchanged, when unconfigured. An unopenable path fails startup closed (`SESSION_STATE_STORE_UNAVAILABLE`), mirroring the existing `audit_store` open-failure pattern exactly.
- Extracted into its own helper rather than inlined, to avoid adding to `run_startup`'s already-high pre-existing complexity (measured at 26 on clean `main` before this change; unchanged after, since the helper is a single call site).

### Added after review

- **Dropped `or None` from `raw.get("session_state_path")`. Set but falsy values (`""`, `false`, `[]`) previously collapsed into the unconfigured default and started the gateway silently without the store; they now fail closed.**
- `config.py`: `session_state_path` is also rejected when it is not a string, or when it is empty or whitespace only. `"   "` previously passed the traversal check and created a file named for the whitespace.
- `docs/spec/error-codes.md`: registered `SESSION_STATE_STORE_UNAVAILABLE`, which the registry requires before a code may be emitted.
- `docs/configuration.md`: the `session_state_path` row now states the non-blank requirement and names the startup abort code, since the registry entry points at that page.

## Scope

The store's own logic (locking, hydrate/reset semantics, cross-process behavior) is untouched - `tests/unit/test_session_state_store.py`'s existing suite already exercises `SessionState.hydrate()` against a real `SqliteSessionStateStore`, including a real `multiprocessing`-based cross-process lock test. This PR only closes the wiring gap between that already-tested mechanism and `run_startup`.

## Testing

- New config tests: the key is accepted; non-string values (`123`, `false`, `[]`) are refused; empty and whitespace-only values are refused.
- New `run_startup` tests: unconfigured stays `None`; configured opens a real store and the file exists on disk, closed in `finally`; a store-open failure aborts startup and logs `SESSION_STATE_STORE_UNAVAILABLE`, asserted via `caplog` rather than the exit code alone. That failure is forced via monkeypatch rather than an OS-level directory trick, since this repo's CI runs both Linux and Windows.
- `pytest tests/unit -q` -> 1991 passed, 6 skipped, one pre-existing unrelated failure (`test_startup_fails_on_unknown_tee_provider_name`, fails identically on clean `main`, a local TPM auto-detection quirk unrelated to this change).
- Ruff and mypy clean on all touched source files.


imran-siddique (dismissed):
Approving at 945351a. Both breaks are real and both are fixed: `SqliteSessionStateStore(` had no call site in `src/` on main, so the shared ratchet #629 shipped was inert, and an unopenable path now stops startup rather than leaving the store unset.

It now conflicts with main in `src/cmcp_runtime/startup.py` only. The kill switch work merged today added its own step 5f at the same place. Resolution: keep main's 5f block (the kill switch store) as is, put your `_open_session_state_store` call after it as 5g, and keep both keyword arguments in the `RuntimeContext` call. I did that merge against 9c1117e locally: 1998 passed, 20 skipped, ruff and mypy clean. Could you rebase by 22 September? I will merge on the rebase.

rajnisht7 (commented):
The implementation looks good and the required behavior is covered.

the commit fd72d83 introduces `.claude/SESSION_CONTRACT.md` which is outside the allowed files listed in the contract itself. Was this file intentionally added as part of the PR?

As it seems to be a session/verification contract rather than part of the runtime/product changes.

Yatsuiii:
Sorry about that. I missed the .claude/SESSION_CONTRACT.md in the final diff. I treated the rebase as a routine change and didn't review the final diff before pushing, which I should have. I've removed it now.

altrudev:
DDC-style boundary check found one small config edge worth closing before merge.

Once `session_state_path` becomes reachable through `_KNOWN_TOP_KEYS`, a non-string YAML value (for example `session_state_path: 123`) reaches:

```python
session_state_path = raw.get("session_state_path") or None
if session_state_path is not None:
    _check_no_traversal("session_state_path", session_state_path)
```

`_check_no_traversal()` passes the value to `PurePosixPath/PureWindowsPath`, so a non-string raises `TypeError` rather than the repository's normal `ConfigError` fail-closed path.

Suggested narrow guard before traversal:

```python
if session_state_path is not None:
    if not isinstance(session_state_path, str):
        raise ConfigError("session_state_path must be a string")
    _check_no_traversal("session_state_path", session_state_path)
```

with a regression such as `session_state_path: 123` -> `ConfigError`.

The core #653 wiring in this PR otherwise matches the documented boundary: configured store is constructed and carried into `RuntimeContext`, unconfigured remains process-local, and an explicitly configured but unavailable store aborts startup rather than silently weakening persistence.

rajnisht7:
Good Catch @altrudev but this is pre-existing on `policy_bundle_path`/`catalog_path`/`audit_db_path` too (verified the same raw TypeError, uncaught by run_startup's except ConfigError). As this PR is adding a new path field via the same pattern, happy to see the guard added here, but would suggest keep this PR focused on #653 wiring and create a separate follow-up issue to fix all four consistently, up to you whether you want it in this diff or as a fast-follow.

the commits [294a76b](https://github.com/agentrust-io/cmcp/pull/661/commits/294a76b563dcf48e575d79821eeff046e82b696c) and [945351a](https://github.com/agentrust-io/cmcp/pull/661/commits/945351a3116e97ed0b24c42837c56d148f8e6b8a) arent signed off, also it isnt rebased

Yatsuiii:
Good catch @altrudev. On scope: on main that key is rejected as unknown, so
the traversal check is unreachable and `session_state_path: 123` fails
closed. Adding it to the known keys is what makes it reachable, so the
TypeError is new for this field even though the underlying gap is the same
one @rajnisht7 points at on the other three.

While looking I found a second one, specific to this field. It was read as
`raw.get(name) or None`, so 0 or false got silently swallowed and the
gateway started without the store. None of the other three do that, so I
have fixed it here rather than deferring it.

For the type check I will open a follow-up covering all four rather than
guard one field here. @altrudev you found it, so that one is yours if you
want it. Otherwise I will take it.

rajnisht7 (changes requested):
Reviewed the commit [88a5939](https://github.com/agentrust-io/cmcp/pull/661/commits/88a5939bb84449bcde5f01beef5e9c7f96719adf), a few things to look upon:

- `SESSION_STATE_STORE_UNAVAILABLE` isn't in `docs/spec/error-codes.md`, as it says that every emitted code must be registered. Kindly add a row (status `startup`, level `FATAL`, ref `configuration.md`)

- `session_state_path: 123` (or `false`, `[]`) now raises a raw `TypeError` instead of `ConfigError`. Can you add an `isinstance(..., str)` check like `agent_manifest.path` has, along with a test. It should also reject empty and whitespace only values there, since `"   "` currently creates a file named `   `.

- The fail-closed startup test only checks `SystemExit(1)`, so it still passes with a wrong event code, it will be good to assert the event via `caplog`

Yatsuiii:
All three done in 4f56d0a.

Registry row added for SESSION_STATE_STORE_UNAVAILABLE (startup, FATAL,
ref configuration.md). I also updated the configuration.md row itself,
since the registry entry points there and that page only documented `..`
rejection.

session_state_path now rejects non-strings and empty or whitespace only
values before the traversal check, so your `"   "` case no longer creates a
file named for the whitespace. Tests are parametrized over 123, false, []
and over "" and "   ".

The fail-closed startup test now asserts the event via caplog rather than
just SystemExit(1). I checked that it fails if the event code is changed,
so it is actually pinning it.

One overlap worth flagging: @altrudev opened #674 for the follow-up and put
the isinstance check inside _check_no_traversal, which covers every path
field including this one. No conflict with this branch, I checked the
merge. My inline check becomes redundant once his lands and I am happy to
drop it on whichever merges second. The empty and whitespace rejection is
not in #674, so that part stays either way.

rajnisht7 (approved):
#653 wiring looks good, config accepts session_state_path, startup opens SqliteSessionStateStore into RuntimeContext, unopenable path fails closed with SESSION_STATE_STORE_UNAVAILABLE, and the new tests cover the important paths as well.

nit: dropping `or None` from `raw.get("session_state_path")` or `None` is good behavior tightening falsy but set values ("", false, []), now fail closed instead of being silently treated as unconfigured. Kindly add a one-line callout in the PR description as well

non-blocking:

- it would be good to have an end to end regression test, that is session_state_path -> run_startup -> use ctx.session_state_store with a fixed session_id -> persist a raised sensitivity -> reopen the same DB path via a fresh startup/store -> assert hydrate adopts the stored value. (Store-level same-id persistence is already covered in test_session_state_store, this would lock the startup wiring path end-to-end)

There is one more gap not introduced by this PR that is restart / cross-instance continuity. The store is keyed by session_id, but SessionManager always generates a new uuid4() on create and has no durable resume/reattach path so a restarted or peer instance cannot pick up the prior logical session’s row. That predates this PR (#629 + the documented “survive restart / across instances” claim). Wiring the store is necessary but not sufficient, worth tracking separately so the session-identity contract is explicit.

Separately (also pre-existing, also not introduced by this PR): session_state_store is never explicitly closed anywhere, same as audit_store and kill_switch_store (both already unclosed on main before this PR). There's an existing shutdown hook (MCPServer._lifespan) that could own this e.g. pass session_state_store into MCPServer alongside audit_chain and close it in the lifespan's finally, after proxy.shutdown(). will be good to fix all three stores together rather than singling this one out, so closing behavior stays consistent across them.

carloshvp (changes requested):
Requesting changes on `4f56d0a41a2787da11b2dc92093ba721518d7b70`.

The config allowlist, startup construction, fail-closed open failure, and `RuntimeContext` wiring are correct in isolation. The production path still cannot hydrate the same logical session after a restart or on another instance, though, so the documented and issue #653 guarantee remains unmet.

`build_server()` calls `SessionManager.create_session()`, and `_open_session()` always assigns a fresh `uuid4()`. There is no production reattach or restore path that supplies the session ID already stored in SQLite. I reproduced this with two fresh `run_startup()` plus `build_server()` calls using the same config and state database:

- process 1 created session `d93eb499-5303-4c9d-8519-1decc253ebe5` and persisted sensitivity `pii`;
- process 2 created session `1a2c002d-6a70-4a18-9d0e-24d8ebe91a5f`, `hydrate()` returned `False`, and sensitivity remained `public`.

A fixed-ID lower-level probe does hydrate `pii`, which confirms the SQLite store works. The missing piece is carrying an authenticated, stable logical session identity into the new process or instance. With this patch, startup now succeeds, so an operator can reasonably believe the advertised restart and horizontal-scaling protection is active when the actual server path still starts a new ratchet.

Please add an end-to-end regression through the production composition path that raises sensitivity, constructs a fresh runtime/server over the same database, resumes the same authorized session, and verifies hydration before policy evaluation. The restoration mechanism must bind the reused identifier to the correct agent/session authority rather than accepting an arbitrary client-selected ID. If that restoration is intentionally outside this PR, the restart and cross-instance claims need to be narrowed and #653 should remain open because its expected behavior is not yet satisfied.

Validation performed:

- reproduced `main` rejecting `session_state_path` as an unknown key;
- 128 focused config/startup/store/session-manager tests passed;
- full CI-equivalent suite: 2,090 passed, 18 skipped;
- `ruff check src/ tests/` passed;
- strict `mypy src/cmcp_runtime/` passed;
- clean `origin/main` merge simulation and `git diff --check`.

Those checks validate the scoped wiring, but none currently exercises session identity continuity across a real restart or second server instance.

imran-siddique:
Raghav, keep this PR focused on connecting the session store. Please update the docs and changelog to say that restarting or sharing a database does not yet restore the same logical session. Remove `Closes #653` and leave that issue open.

Restoring a session needs a separate change that checks who may resume it, with a test through the real startup path. Once the claims are corrected, we can re-review this smaller fix.

imran-siddique:
Corrected the docs and changelog to limit this PR to opening and connecting the store. Restarting or sharing a database does not restore the same logical session. Removed the closing keyword so #653 stays open for authenticated resumption. All 99 configuration, startup and session-store tests pass locally.

imran-siddique (dismissed):
Re-reviewed the implementation as a startup-wiring change. Configuration rejects invalid and blank paths, configured stores open through run_startup, and open failures abort with the registered error code. All 99 config/startup/store tests pass locally. My follow-up only narrows the docs and changelog; authenticated session restoration remains in #653, which this PR no longer closes.

rajnisht7 (dismissed):
Verified the full wiring for #653, config now accepts session_state_path with proper string/non-blank/traversal checks, startup opens the store (or leaves it None) and fails closed with SESSION_STATE_STORE_UNAVAILABLE if it cant and the store is passed into RuntimeContext

non-blocking:

- stale inline documentation for Config.session_state_path as it still says configuring the path makes the ratchet survive restart/across instances, while the corrected docs now explicitly say that logical-session resumption is not implemented yet.

- import style: from pathlib import Path as _Path inside the helper is fine, but Path is already used a few lines above for the kill-switch store. A top-level import would have been cleaner.

Yatsuiii:
Thanks for updating the docs and changelog and re-reviewing this, Imran. I’ve caught up and agree with the narrowed scope.

imran-siddique:
@Yatsuiii please align the `Config.session_state_path` comment with the corrected docs: configuring the store does not yet restore the same logical session after restart or across instances. Authenticated resumption remains in #653.

@carloshvp the PR now follows your narrower-scope option, and #653 remains open. Once that last comment is corrected, could you re-review and resolve your change request if satisfied?

carloshvp (changes requested):
The technical blocker from my earlier review is resolved on a1b5b089. The changelog, configuration reference, and Config comment now correctly limit this PR to opening and wiring the store, and leave authenticated restart/cross-instance resumption in #653. I reran the 99 focused config, startup, and session-store tests; Ruff, strict mypy, Bandit, diff checks, and a merge simulation against current main all pass.

One repository requirement remains: the latest commit a1b5b089 has no Signed-off-by trailer, while CONTRIBUTING.md requires one on every commit. Please amend/sign off that commit and force-push. I will re-check the resulting head.

Yatsuiii:
Could you rerun the failed Windows 3.13 and linux-docker jobs? Windows passed all tests and failed only during Codecov's GPG key import. I couldn't reproduce the linux-docker failure from the CI snapshot locally, so I'd like to see whether it reproduces before changing anything.

rajnisht7:
@Yatsuiii Thank you for checking this, re-ran both jobs and they're green now. The linux-docker failure was in `test_supervision.py::test_independent_lease_stops_container_after_bridge_failure[watchdog-pause-unguarded]` (1 failed, 60 passed), it failed at its very first check, before the test injects any fault, and as that code doesnt touch this PR (the confinement tests never call load_config or run_startup), it appears like a flaky timing issue in that test, so nothing to change on your side.

Issue #659, opened 2026-09-18
Validate that agent egress cannot bypass configured release controls


imran-siddique:
Parent tracker: https://github.com/agentrust-io/.github/issues/41

## Problem and scope

Merged #657 supplies operator-owned tool and response sensitivity ceilings. Those checks do not mediate direct agent sockets, files, subprocesses, alternate model endpoints or every log sink.

Define a reusable confinement contract and exercise one supported reference deployment. Keep OS/runtime adapters explicit; do not claim that an SDK wrapper alone confines arbitrary agent code.

## Acceptance criteria

- [x] Inventory plaintext sources and sinks, subprocess inheritance, local files, DNS/network, alternate endpoints, stdout/stderr, telemetry and audit metadata.
- [x] Define which component mediates each route, its trusted boundary, startup behavior and behavior when the mediator is unavailable.
- [x] Build a canary harness with permitted-path positives and direct bypass attempts. Observe attempted release at independently controlled sinks, not only application error messages.
- [x] A chosen reference sandbox blocks routes outside the approved gateway; removing its relevant restriction makes the corresponding test fail.
- [x] Test restart, child processes and configuration changes without a fail-open window.
- [x] Publish supported deployment assumptions and untested channels, including covert/side channels. Canary scans do not establish absence of all leakage.

## Dependencies and boundaries

#124 owns policy enforcement inside the TEE; #462 owns measurement-bound signing-key custody. This issue owns mediation around that enforcement. Hardware-backed claims require those dependencies and measured application identity. Existing #657 is merged; do not reopen its completed scope or treat its unit tests as confinement evidence.
## Reference implementation status (2026-09-18)

The checked criteria refer to the bounded native-Linux reference, with acceptance review still pending. #662 was approved and merged. Follow-up #663 at `a56f5ab` adds bridge/watchdog failure handling, live operator restrictions and audit/export probes.

The [hosted confinement run](https://github.com/agentrust-io/cmcp/actions/runs/35372651855) passed all 57 tests in 22.34 seconds; all 17 hosted checks passed, with two conditional skips. The `confinement-evidence` artifact contains independently observed delivery/termination outcomes and host versions. The full local suite passed 1,992 tests with 25 skips and 89.46% coverage.

[...]

PR #663, opened 2026-09-18
Supervise confined agents and validate emitted enforcement claims

### LINKED ISSUE #654: _build_policy silently signs unrecognised enforcement_mode values as "advisory" in TRACE claims
## cmcp-gateway version

## Python version

## TEE provider
N/A

## Reproduction steps
1. Construct a PolicyBundleInfo with enforcement_mode set to a value outside {enforcing, advisory, silent}, e.g. "enforce", "declared", "audit-only", "ENFORCING", or "".
2. Build a claim via _build_policy.
3. Inspect the emitted claim's enforcement_mode and validate against schemas/trace-claim.schema.json.

## Expected behavior
An unrecognised enforcement mode is refused at claim-construction time (mirroring the sibling _build_runtime and the merged AUDIT-003 precedent in the same file), so a claim cannot assert a policy posture that was never evaluated.

## Actual behavior
`_build_policy` (src/cmcp_runtime/audit/trace_claim.py:380-386) maps via `mode_map.get(bundle.enforcement_mode, "advisory")`. The map has only enforcing/advisory/silent, and `enforcement_mode` is an unvalidated str, so every other value, including "enforce" (a value this same file emits), "declared" (the trace-spec vocabulary for "nothing evaluated it"), "audit-only", "ENFORCING", and "", is silently signed as advisory. Per TRACE v0.2 section 4.3, enforce/advisory/silent all positively assert that *something evaluated the policy*, so this emits a false attestation.

Reproduced on 4ca6102: six unrecognised inputs all became signed `advisory`, and every resulting claim passed schemas/trace-claim.schema.json.

All six unrecognised inputs → signed enforcement_mode "advisory", schema-valid.

TRACE (claim integrity; relates to TRACE v0.2 section 4.3, docs/SPEC.md:204)

## Additional context
In-repo precedent for refusing rather than defaulting:
- AUDIT-003, trace_claim.py:343-351 (test tests/unit/test_trace_claim.py:425-426): "unknown attestation provider must be rejected, not silently accepted" / "Rejecting claim construction to prevent spoofed attestation reports."
- policy/decisions.py:149-150: anything unrecognised becomes deny rather than passing through.
- trace-spec merged PR #143 argues "declared" is never a default: https://github.com/agentrust-io/trace-spec/pull/143
- LIMITATIONS.md:80 argues for refusing to sign rather than signing an untrue record.

imran-siddique:
A confined agent must stop when supervision fails, and emitted TRACE claims must describe valid enforcement modes and validate against their published schema.

## Changes

- Require bounded watchdog acknowledgments and stop the agent after bridge or watcher failure, under the documented surviving-host assumptions.
- Serialize live operator restrictions with dispatch while preserving session classification. Observe actual audit, log and telemetry exports without private exception text.
- Refuse unrecognized enforcement modes instead of silently signing them as advisory. The valid mappings and gateway enum path are unchanged. This incorporates #670.
- Declare the emitted `gateway.call_log_summary` and call-graph `edges_represent` fields in the TRACE schema. Keep them optional for older claims and reject malformed values and undeclared properties. Validate complete HTTP session-close and operator-trip claims, with declaration-removal controls.

## Validation

At `40e3b2d`, all six Python/OS test jobs, native Docker supervision, governance, lint/type/security checks and signing compatibility passed. Fuzzing also passed; no failing or unfinished checks remain. The Linux Python 3.12 job reports 2,101 passed and 29 skipped, including both new HTTP schema tests.

The two retained live claims from the kill-switch paper each fail the previous schema for exactly these two undeclared fields and pass the corrected schema. The 45 existing claim tests, Ruff and mypy passed locally. Windows application control blocked the local Cedar module for the new HTTP tests; those tests passed in hosted CI. Historical paper results remain unchanged.

Prior bounded-ack source `bad751b` passed 61 native tests in [run 35445862949](https://github.com/agentrust-io/cmcp/actions/runs/35445862949), including the paused watcher and deliberately weakened supervision. The six invalid-mode controls failed before the enforcement-mode correction.

## Limits and review

The host, Docker daemon and surviving supervisor remain trusted. Simultaneous bridge/watcher loss, host-wide orphan recovery, arbitrary exporter plugins and secure erasure remain outside this reference. Policy revocation does not undo dispatched effects. The watchdog review concern still requires independent re-review; this update does not dismiss it. Auto-merge remains disabled.

Continues #659. Closes #654. Enforcement-mode reproduction credit: @saintmalik. Original histories, review discussions and DCO sign-offs are preserved.


carloshvp (changes requested):
Blocking: `LeaseWatchdog.pulse()` treats a successful write/drain to the child stdin pipe as proof that the watchdog is alive. That is not a liveness acknowledgement: the pipe continues buffering heartbeats while the watchdog is hung or stopped. I suspended the watchdog after its `ready` line, waited 3.25 seconds (past the 2-second lease), and observed `watchdog_returncode=None` and `pulse_done=False`. The bridge therefore remains active with no responsive supervisor, and if the bridge subsequently dies the stopped watchdog cannot process EOF or stop the plaintext-bearing container.

This contradicts the documented claim that watchdog failure terminates the bridge exchange and leaves a fail-open supervision state untested. Please require a bounded acknowledgement from the watchdog (or another independent health signal) so missed watcher progress closes the bridge and invokes its own container-stop path, and add a causal Linux case that pauses the watchdog while the bridge remains active.

Qiang-Xu (dismissed):
Looks good to me over all.

Qiang-Xu:
> Blocking: `LeaseWatchdog.pulse()` treats a successful write/drain to the child stdin pipe as proof that the watchdog is alive. That is not a liveness acknowledgement: the pipe continues buffering heartbeats while the watchdog is hung or stopped. I suspended the watchdog after its `ready` line, waited 3.25 seconds (past the 2-second lease), and observed `watchdog_returncode=None` and `pulse_done=False`. The bridge therefore remains active with no responsive supervisor, and if the bridge subsequently dies the stopped watchdog cannot process EOF or stop the plaintext-bearing container.

That is a valid concern, we may want to check the buffer size, @imran-siddique thoughts?

imran-siddique:
@Qiang-Xu buffer size would not establish that the watchdog is responsive. A write can succeed while the watchdog is stopped.

`bad751b` now requires a watchdog acknowledgement within one second. A missing or invalid reply ends the exchange. It also adds the paused-watchdog test and a control with supervision disabled.

The [Linux run](https://github.com/agentrust-io/cmcp/actions/runs/35445862949) passed all 61 tests at that commit, including both new cases.

@carloshvp could you re-review the fix against your blocking finding?

imran-siddique:
@Qiang-Xu could you review the combined changes at `7573b36796db`? This now includes #670 alongside the confinement work. Please check the bounded watchdog acknowledgement repair and the refusal of invalid enforcement modes. The title and description cover the combined scope and validation; the existing watchdog review concern remains for re-review.

imran-siddique:
@zohebk8s @Qiang-Xu could you review the current head, `40e3b2d`? It includes the bounded watchdog acknowledgement, invalid-enforcement-mode refusal, and the schema correction for emitted audit summaries. All current CI checks passed; the Linux Python 3.12 job reports 2,101 passed and 29 skipped.

@carloshvp could you also re-review the paused-watchdog finding against the bounded-ack fix? The existing blocking review remains open. The PR description links the validation and states the surviving-host assumptions.

carloshvp (approved):
Validated current head 40e3b2d719dc411361e77191310ce61fc0f37d9a. The prior blocking liveness finding is resolved: every heartbeat now requires a bounded one-byte watchdog acknowledgement; missing, malformed, paused, or killed watchdog paths fail the bridge and trigger container stop.

Independent validation: real-process SIGSTOP and SIGKILL cases passed; 87 focused tests passed; the full local suite passed with 2,105 passed and 25 platform-dependent skips; Ruff, strict mypy, Bandit, diff check, and a clean merge simulation passed. The native Docker lifecycle cases are platform-skipped locally, and the current hosted linux-docker job is green, including the causal paused-watcher and weakened-ack controls. I also checked that invalid enforcement modes fail closed and complete emitted claims validate against the schema.

The documented surviving-host and Docker-daemon limits are explicit and consistent with the implementation. Approving this exact head.

zohebk8s (approved):

PR #665, opened 2026-09-18
fix: keep kill switch blocks across restarts and stop cleanly when one trips

imran-siddique:
Three defects in the kill switch, each reproduced on main (49392a4) before the fix:

- Blocked identities lived in an in-memory set (`kill_switch.py:33`), so any restart lifted every block. Measured: after a trip, a fresh `build_server` started with the identity unblocked.
- A close that tripped the switch raised `KillSwitchTripped` while creating the successor session (`mcp/server.py`, `_session_close`). The caller got `500 INTERNAL_ERROR` instead of the signed claim, and the admission barrier stayed sealed. Measured: the next `tools/call` got no answer within 3 seconds instead of the documented `403 KILL_SWITCH_TRIPPED`.
- The tutorial documented an unblock at `DELETE /admin/kill-switch/{id}` that does not exist, and said the unblock was logged to the audit chain. `KillSwitchEvaluator.unblock()` had no caller.

What changes:

- Blocks are stored in a `kill_switch_blocks` table in the audit database, which startup already opens and refuses to run without. Reads hit the database each time, so instances sharing the file see each other's blocks. The rolling window is not stored: a restart can delay a trip, never lift one.
- A tripping close returns its claim and halts the gateway. Calls get `403 KILL_SWITCH_TRIPPED` at once, close and reset get `409`, and `/readyz` reports a `kill_switch` failure.
- A gateway whose identity is already blocked starts halted instead of raising out of `build_server`, and records `kill_switch_block_active_at_start` as the first entry after `session_start`.
- `POST /kill-switch/unblock` is an operator route. It needs `agent_id`, `reason` and `authorized_by`, records a `break_glass_used` entry with the credential that was verified, and resumes on a new session. A retry after a failed resume still resumes.

Tested with the CI install (`requirements/dev.txt` with hashes, then `-e .`) on Windows, Python 3.12: `pytest tests/unit tests/conformance tests/integration tests/confinement` 1946 passed, 20 skipped; `tests/unit/test_kill_switch_durable.py` adds 14. `ruff` and `mypy src/cmcp_runtime/` are clean. Two existing fixtures now set `halted_identity = None` on their `MagicMock` proxies.


Qiang-Xu (dismissed):
Looks good to me!

PR #666, opened 2026-09-18
fix: stop a session at the call that trips the kill switch

imran-siddique:
Stacked on #665. The deny rate was evaluated only in `close_session`, so an agent the kill switch should have stopped kept calling for as long as its client held the session open. A client that never closed was never stopped.

What changes:

- Each call is counted as it completes, by the decision in its audit entry, using the rule the close-time count already used: `allow` is allowed, `deny` and `advisory_deny` are denied, anything else is not counted.
- At the call that crosses the threshold the gateway refuses further admission at once, closes the session, signs its claim with `kill_switch_triggered: true`, and appends `break_glass_used` with `tripping_call_id`. The claim is at `GET /sessions/{id}/trace-claim`.
- Calls counted mid-session are tracked on the session and not counted again at close.
- `POST /kill-switch/trip` lets an operator stop the bound identity without waiting for its deny rate. It needs `reason` and `authorized_by`, blocks the identity, records the trip and the credential, and returns the closed session's claim. It returns `409` with no Agent Manifest bound (`NO_AGENT_IDENTITY`) or with the switch disabled (`KILL_SWITCH_DISABLED`).

One limit, added to LIMITATIONS.md: a trip stops new calls immediately, but calls admitted before it are drained as on any close. They finish, or are cancelled at `CMCP_SESSION_CLOSE_DRAIN_SECONDS`.

Tested the same way: 1955 passed, 20 skipped. `tests/unit/test_kill_switch_mid_session.py` adds 9, driving real admission, observation, seal and halt with a denying enforcement stub.


Qiang-Xu (dismissed):
LGTM.

PR #667, opened 2026-09-18
fix: make kill switch state and refusals checkable by a verifier

imran-siddique:
Stacked on #666. Two gaps in what a verifier could establish, and one configuration that looked armed and was not:

- A claim with `kill_switch_triggered: false` read the same whether the switch was enabled and did not trip, or was never enabled.
- A call refused by a tripped gateway produced nothing a verifier could check.
- `kill_switch.enabled: true` with no Agent Manifest started normally. The tutorial said anonymous sessions are never blocked, so the switch could stop nothing.

What changes:

- Claims from a gateway with the switch enabled carry `gateway.kill_switch`: `enabled`, `window_seconds`, `deny_rate_threshold`, `min_calls`, and `trigger` (`deny_rate` or `operator`) when it tripped. The field is omitted otherwise, so other gateways' claims are byte-identical. The schema addition is optional.
- Each refused call returns a receipt in `error.data.receipt`, type `cmcp.kill-switch.refusal/v1`, signed with the claim signing key over the claim's own `canonical_json`, and naming the closed session's claim by digest.
- `cmcp_verify.verify_kill_switch_refusal(receipt, claim)` checks the claim signature, that the receipt was signed by the same key, the digest, the session, and that the claim records a trip. It reports every failure. It does not establish the hardware binding: that is `verify_trace_claim`'s job, and the module docstring says to run both.
- Startup aborts with `KILL_SWITCH_REQUIRES_IDENTITY` when the switch is enabled without a manifest.
- The tutorial drops "verifiable offline by any regulator" and "No other MCP gateway can produce this artifact", replaces the sovereign section with what the evidence does and does not show, and is renamed from "AGT SRE Kill Switch".

Found while testing, not fixed here: the published claim schema already rejects claims the gateway emits on HTTP close (`gateway.call_log_summary` and `call_graph_summary.edges_represent` are not in it). Reproduces on main with the kill switch off. These tests validate only the new `kill_switch` object.

Tested the same way: 1968 passed, 20 skipped; `tests/unit/test_kill_switch_evidence.py` adds 13, including altered receipts, a receipt against another gateway's claim, and a tampered claim. `mkdocs build --strict` passes.


Qiang-Xu (dismissed):
Looks good.

PR #668, opened 2026-09-18
fix(policy): revoke a compromised policy signing key without a restart

imran-siddique:
A gateway with hot reload on trusted `CMCP_POLICY_SIGNING_KEY` until someone changed the config and restarted it. STATUS.md:39 and docs/configuration.md:131 said so. Until each gateway restarted, whoever held a stolen key could keep installing signed bundles with higher versions on it.

While testing the fix I found a second defect underneath it. `PolicyStore.reload_if_stale` (src/cmcp_runtime/policy/bundle.py, the `load_policy_bundle(self._bundle_path, self._expected_hash, self._signing_key)` call on main) passed the startup `CMCP_POLICY_HASH` to every reload. Outside dev mode that hash is required, so with a hash and a key both pinned, the shape configuration.md calls the supported production one, a correctly signed newer bundle failed with `PolicyHashMismatch` and reload never installed anything. Measured before the fix: reload returned False and version 1.0.0 stayed in force. The first commit fixes that: with a key pinned, the key authorises reloads and the hash is still checked on the startup load. Recovery after a revocation depends on it.

What changes:

- `CMCP_POLICY_SUCCESSOR_SIGNING_KEY` pins a second Ed25519 public key at startup. While the current key is trusted, the successor signs no bundles.
- `signing-key-revocations.json` in the bundle directory holds statements of the form `{"revoked_key_id": "sha256:<hex of raw key>", "signature": "<b64url>"}`. The signature covers `cmcp-policy-key-revocation-v1|<id>`, a different domain from bundle signatures. The file sits outside the bundle hash.
- Each reload attempt applies the file before it reads the bundle, and so does startup, before the first load. Only the current key can be revoked. The successor or the current key itself may sign the statement. Revoking the current key promotes the successor.
- The policy in force was signed by the revoked key, so the gateway fails closed. `PolicyEvaluator.evaluate` calls `PolicyStore.require_trusted()`, which raises `PolicySigningKeyRevoked` (code `POLICY_SIGNING_KEY_REVOKED`, a `PolicyDeny` subclass) in enforce, advisory and silent modes until a successor-signed bundle with a higher version is installed. The proxy records each refused call as a deny in the audit chain. A bundle signed by a revoked key is refused on reload and at startup with the same code.
- Revocation cannot be undone within the process. No statement un-revokes a key. Deleting the file changes nothing at runtime, replaying a statement already applied does nothing, and a stolen current key cannot revoke the successor. Malformed or wrongly signed statements are logged as `POLICY_KEY_REVOCATION_INVALID` and skipped. They do not block other statements or the reload itself, because anyone who can write the directory can write junk.
- TRACE claims gain optional `gateway.policy_signing` with `key_id` (the key the policy in force verified under) and `revoked_key_ids`. The schema change is additive. With no key pinned the field is absent and claims are byte-identical.
- Startup now turns `PolicySignatureInvalid` into a FATAL entry and exit 1. On main the exception propagated out of `run_startup`.

Why a pre-pinned successor plus a signed statement, not a key list in config: a list read from the environment still needs a restart to change, which is the defect. A statement signed only by the current key would put defender and thief on equal footing. The successor's private key is held apart, so its signature is the one thing the thief lacks. Delivering the statement through the bundle directory reuses the existing reload poll and adds no control endpoint.

Out of scope, and stated in STATUS.md, LIMITATIONS.md and the hot-reload spec. There is one successor per startup, so a second compromise before the next restart can only be answered by self-revocation, which refuses every call. Revocation state is in memory: a restart rebuilds it from the environment plus the file, so deleting the file and keeping the old key configured re-trusts it. Nothing distributes statements to a fleet, keys never expire, and with reload off a revocation takes effect only at the next startup.

Tested on Windows, Python 3.12.10, fresh venv installed the CI way (`pip install --require-hashes -r requirements/dev.txt` then `pip install --no-deps -e .`):

- `pytest tests/unit/ tests/conformance/ tests/integration/ tests/confinement/`: 1957 passed, 20 skipped
- `tests/unit/test_policy_key_revocation.py`: 24 new tests. Rotation to the successor, self-revocation, refusing a newer bundle signed by the revoked key, fail closed in all three modes through a real Cedar backend, a deny recorded through `CMCPProxy.call_tool`, no un-revoke, no revoking the successor, replay as no-op, bad statements skipped, startup applying and refusing, and the claim validating against the schema
- `tests/unit/test_policy_signing.py`: 2 new tests for both pins
- Mutation checks: with the `require_trusted()` call removed, 4 revocation tests fail. With the old hash re-check restored, the both-pins test fails
- `ruff check src/ tests/`, `mypy src/cmcp_runtime/`, `bandit -r src/ -c pyproject.toml`, `mkdocs build --strict` and `scripts/check_docs_quickstart.py` are clean. `pip-audit` was not run, because no dependencies changed


Qiang-Xu (approved):
Looks good, thanks!

PR #670, opened 2026-09-18
fix(claim): refuse an unrecognised enforcement mode instead of signing advisory

### LINKED ISSUE #654: _build_policy silently signs unrecognised enforcement_mode values as "advisory" in TRACE claims

imran-siddique:
On main, `_build_policy` (`trace_claim.py:432`) mapped any enforcement mode outside `enforcing`, `advisory` and `silent` to `advisory` and signed it. Each of those three values says the policy was checked, so a claim built from `enforce`, `declared`, `ENFORCING` or an empty string carried an evaluation that never happened. The sibling `_build_runtime` in the same file already refuses an unknown provider (AUDIT-003).

Scope: the gateway's own path passes `str(ctx.config.attestation.enforcement_mode)`, and `EnforcementMode` is a `StrEnum`, so it always yields one of the three valid values. The fallback was reachable only by code that builds claims directly.

What changes: `_build_policy` raises `ValueError` for an unrecognised mode, with the same wording shape as AUDIT-003. The three valid modes map as before.

Tests: the six inputs from #654 (`enforce`, `declared`, `audit-only`, `ENFORCING`, empty, `Enforcing`) each raise, and the three valid modes map unchanged. Against main's `trace_claim.py` the six fail; with the fix all 11 pass. Full suite: 2003 passed, 20 skipped; ruff and mypy clean.

Reported by @saintmalik, who traced it to the line and supplied the reproduction inputs.

Closes #654


Qiang-Xu (dismissed):
Looks good.

rajnisht7 (approved):
Looks good, `_build_policy` now raises `ValueError` for any mode which is outside enforcing/advisory/silent matching `_build_runtime`'s provider refusal. The three valid mappings remain unchanged, and the tests cover the six #654 inputs. The gateway path passes a validated StrEnum, so it's also unaffected, only direct callers could hit the old fallback. 
Non-blocking: no test for `None`/`non-str` input or through `generate_trace_claim`, and `_ENFORCEMENT_MODE_MAP` could be annotated like `_PROVIDER_MAP`

imran-siddique:
Consolidated into #663 at 7573b36796db98ed8e3acf729a0712700b960b01. The combined branch preserves this PR's complete head and history, and its title/description now cover both changes. Closing this duplicate as superseded; this is not a merge or a resolution of outstanding review requirements.

PR #671, opened 2026-09-18
fix(catalog): refuse tools whose names derive the same Cedar action

### LINKED ISSUE #655: Catalog accepts distinct tools that derive the same Cedar action, collapsing them to one policy identity
## cmcp-gateway version

## Python version

## TEE provider
N/A

## Reproduction steps
1. Define two distinct tools whose names differ only by underscores/leading/trailing underscores, e.g. `read_file` and `read__file` (also `_read_file`, `read_file_`).
2. Load them via load_catalog and run policy in ENFORCING with a real PolicyBundle.
3. Write an action-scoped permit/forbid for one tool and observe it applies to the other.

## Expected behavior
Catalog load fails closed on the collision (raise CATALOG_TOOL_NAME_COLLISION), consistent with docs/spec/tool-identity.md section 3 and Issue #86's acceptance criterion — two distinct tools must never share one policy identity.

## Actual behavior
CedarBackend.build_request derives the Cedar action from the tool name via `"".join(part.capitalize() for part in tool_name.split("_"))` (src/cmcp_runtime/policy/cedar.py:54-58). This is not injective on lowercase input: read_file, read__file, _read_file, and read_file_ all map to Action::"ReadFile". load_catalog guards literal lowercase and exact duplicates but does not check the *derived* action, so two tools with different server identities share one policy identity; an action-scoped permit/forbid for one applies to the other. This is the unfixed half of the case-collision defect closed fail-closed by #159 / PR #218.

Live repro with real load_catalog + real PolicyBundle in ENFORCING; the workflow-scoped permit idiom from tests/unit/test_egress_policy.py:399-401 authorizes both colliding tools.

POLICY (tool identity / catalog load)

## Additional context
- docs/spec/tool-identity.md section 3: "The runtime does not attempt to resolve the collision. It fails closed" / "Silent collision resolution … would allow a malicious or misconfigured server to shadow a legitimate tool."
- Issue #86 acceptance criterion: "Detects tool_name collisions at load time; raises CATALOG_TOOL_NAME_COLLISION (gateway refuses to start)."
- Issue #159 / merged PR #218 already rejected the case half at load.

imran-siddique:
The Cedar backend names a call's action by joining the underscore-separated parts of the tool name, each capitalised (`cedar.py:55` on main). That is not injective: `read_file` and `read__file`, for one pair, both give `Action::"ReadFile"`. The catalog loader refused literal duplicates and uppercase names only, so two such entries loaded, shared one policy identity, and a permit or forbid for one applied to the other. `docs/spec/tool-identity.md` §3 already requires a collision to fail closed; this is the half #218 did not reach.

What changes:
- The derivation moves to `cmcp_runtime.policy.action_name.cedar_action_name`, used by both `CedarBackend.build_request` and `load_catalog`, so the check cannot drift from what the backend evaluates.
- `load_catalog` raises `CatalogToolNameCollision` when two entries derive the same action, naming both tool names and the action.
- The mapping itself is unchanged, so existing policies keep their action names.
- §3 of the tool-identity spec gains a paragraph for the derived-action case.

Tests: four names that collide with `read_file`, including a triple underscore not in the report, each refuse to load and fail against main's loader; `read_file` with `readfile` still loads; and a check that the backend and the loader derive the same action. Catalog tests 35 passed; full suite 2000 passed, 20 skipped; ruff, mypy and `mkdocs build --strict` clean.

Reported by @saintmalik, who reproduced it with a real catalog and bundle in enforcing mode.

Closes #655


Qiang-Xu (approved):
LGTM.

PR #674, opened 2026-09-19
fix(config): classify non-string path settings

### LINKED ISSUE #673: Path config settings crash with TypeError instead of ConfigError on a non-string value
Split out of #661. @altrudev found the missing type check on the new
`session_state_path` field there, and @rajnisht7 verified the same
behaviour on the three fields below and suggested tracking it separately.

`policy_bundle_path`, `catalog_path` and `audit_db_path` are passed
straight to `_check_no_traversal()` with no type check
(`config.py:495-497`). A non-string YAML value reaches `PurePosixPath()`
at `config.py:263` and raises `TypeError`. `run_startup()` only catches
`ConfigError` around `load_config()` (`startup.py:361-364`), so this
surfaces as an unhandled traceback instead of the normal fail-closed
`CONFIG_ERROR` exit.

Repro on main (9db65b5). Same result for all three:

```yaml
policy_bundle_path: 123
```

```
run_startup -> UNCAUGHT TypeError: argument should be a str or an
os.PathLike object where __fspath__ returns a str, not 'int'
```

Expected: `ConfigError`, so it exits through the normal `CONFIG_ERROR` path.

The pattern already exists in the same function. `agent_manifest.path` and
`agent_manifest.trust_anchor_path` get an `isinstance(..., str)` check at
`config.py:502-505` before their traversal calls at `config.py:515-517`.
The three top-level path settings skipped it.

`session_state_path` will need the same guard once #661 merges. On main it
is rejected as an unknown key, so it is not affected yet.

altrudev:
Closes #673.

The path traversal helper assumed callers had already supplied a string. Three currently reachable top-level config fields — `policy_bundle_path`, `catalog_path`, and `audit_db_path` — pass YAML values straight into that helper, so values such as integers, booleans, lists, or mappings reached `PurePosixPath` / `PureWindowsPath` and leaked a raw `TypeError` instead of the normal fail-closed `ConfigError` path.

This fixes the invariant at the shared boundary rather than adding three field-specific guards:

- `_check_no_traversal()` now rejects non-string values with `ConfigError("<field> must be a string")` before constructing either path object.
- Regression coverage exercises all three reachable top-level path fields across integer, boolean, list, and mapping YAML values.
- Existing traversal behavior for valid strings is unchanged.

This also protects future path fields using the same helper. In particular, once #661 makes `session_state_path` reachable on main, its non-string values will be classified consistently without another one-off guard.

Branch is based exactly on current main `9db65b5`. The diff is limited to `src/cmcp_runtime/config.py` and `tests/unit/test_config.py`.

AI-assisted; Altru.dev remains responsible for the change and verification.

Signed-off-by: altrudev <266135212+altrudev@users.noreply.github.com>


rajnisht7 (approved):
Reviewed the head [a0353fe](https://github.com/agentrust-io/cmcp/pull/674/commits/a0353fe96f0e5dbfac090a38158b266c1af082aa). The type check in _check_no_traversal is the right place for this, tested with odd YAML types (null, float, date, binary, set, NaN, list, dict) through every config field. Only the three path fields leaked a raw exception on base, and none of them do now, also checked locally: ruff, mypy, bandit and the full unit suite all pass. The new tests fail on the base commit (12 failures) and pass with the fix

small non-blocking points:

- The type hint still says path_str: str maybe change it to object/Any later for accuracy.
- Tiny message inconsistency: type error doesn’t quote the field name the way traversal errors do
- Kindly add a short CHANGELOG entry, since bad config now raises ConfigError instead of TypeError
- Consider a null case in the test params. It is now a clean ConfigError (it was a raw TypeError before), while agent_manifest.* treats null as unset. Fail-closed is fine, a test would just pin it.
- The description says #661 will need no extra guard for session_state_path, but #661 adds its own type check. The or None on that line is unreachable here (the key is unknown until #661) and #661 removes it, so there is no real gap, just a wording thing.

imran-siddique (approved):
Approving at `a0353fe`. Fixing the invariant in `_check_no_traversal()` is the right boundary: it covers the three reachable fields today and `session_state_path` when #661 lands, without a fourth one-off guard. Thanks @rajnisht7 for checking the base commit rather than the head alone.

